PR #8, opened 2014-08-21
Prevent crash from ruby on spline generation if empty splines are given

goldhoorn:
The Task inspector seems to pass a empty structure first which should be
filled later. This commit prevents a crash there by passing a empty
spline here.


goldhoorn:
fixed by: https://github.com/rock-core/gui-vizkit/pull/34

PR #17, opened 2014-10-09
Initialize Angle, Pressure and Temperature with base::unknown in default constructor

jakobs:
This is to prevent accidental use of uninitialized values.


doudou:
Since nobody really pushed against it, what about we merge ?

doudou:
Oh well ...

PR #22, opened 2014-12-10
Fix shadow warnings

marvin2k:
Some simple fixes which accumulated while playing around. The most controversial would be dcc0bb005e1ca20253c437dd16709327eb60d339

In 61b4abdefcfe24c6e39277789bfef46a7797d0a3 the default constructor still does zero-initialization -- instead of NaN as discussed on the ML.

### DIFF
--- a/base/Angle.hpp
+++ b/base/Angle.hpp
@@ -264,7 +264,7 @@ public:
     {
     }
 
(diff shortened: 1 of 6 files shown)


doudou:
Since nobody seems to care, I say: we merge.

PR #23, opened 2014-12-18
Minor fixes -- dont use assert

marvin2k:
fixes separated from #22

the change in the exception can be discussed: Only changing this exception but not changing all the others which use the same style...

### DIFF
--- a/base/JointTransform.hpp
+++ b/base/JointTransform.hpp
@@ -33,13 +33,23 @@ struct JointTransformVector : public base::NamedVector<JointTransform>
      */
     void setRigidBodyStates( const base::samples::Joints& joints, std::vector<base::samples::RigidBodyState>& rbs ) const
     {
-	assert( !joints.names.empty() );
+        if (joints.names.empty()) {
+            throw std::runtime_error("base::JointTransformVector::"
+                                     "setRigidBodyStates(): the vector "
+                                     "'joints.names()' is empty");
+        }
 
 	rbs.resize( joints.size() );
 	for( size_t i=0; i<joints.size(); i++ )
 	{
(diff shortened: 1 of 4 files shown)


marvin2k:
removed "rework of exception", rebased to current master, removed even more asserts.

base/Time.hpp
```diff
@@ -96,8 +95,9 @@ namespace base
                 case Microseconds:
                     sprintf(buffer,"%s:%06d", time, uSecs);
                     break;
-                default: 
-                    assert(-1);
+                default:
+                    throw std::runtime_error("base::Time::toString(): invalid "
```
doudou: should probably use std::invalid_argument instead of runtime_error here

base/JointTransform.hpp
```diff
@@ -33,13 +33,24 @@ struct JointTransformVector : public base::NamedVector<JointTransform>
      */
     void setRigidBodyStates( const base::samples::Joints& joints, std::vector<base::samples::RigidBodyState>& rbs ) const
     {
-	assert( !joints.names.empty() );
+        if (joints.names.empty()) {
+            std::stringstream ss;
+            ss << "base::JointTransformVector::setRigidBodyStates(): "
+                  "the vector 'joints.names()' is empty";
+            throw std::runtime_error(ss.str());
```
doudou: why using a stringstream ??? The string is a constant

marvin2k: Hm, the intention was to be able to add the current values of some variables into the exception. Will remove it.

marvin2k:
removed unnecessary `std::stringstream`, used `std::invalid_argument` where appropriate.

PR #24, opened 2015-01-09
bug fix: ruby uses Types::Eigen::Vector3, but should be Eigen::Vector3

hemker:
Wrong namespace

### DIFF
--- a/bindings/ruby/lib/base/geometry/spline.rb
+++ b/bindings/ruby/lib/base/geometry/spline.rb
@@ -408,9 +408,9 @@ module Base
 		    return nil
 		end
                 result = super
-                p = Eigen::Vector3.new(*result[0, 3])
+                p = ::Eigen::Vector3.new(*result[0, 3])
                 if with_tangent
-                    t = Eigen::Vector3.new(*result[3, 3])
+                    t = ::Eigen::Vector3.new(*result[3, 3])
                     return p, t
                 else
                     return p


doudou:
The two names are

```
Eigen::Vector3
```

and 

```
Types::Eigen::Vector3d
```

The reason the ruby type is not named 3d is that the ruby bindings only support double. However, in hindsight, it is quite confusing. Is that your specific problem ? Can you give more context ?

hemker:
Hey,

I found this bug with Janosch; we have a test script for testing the 
local planner in the artemis software stack isolated with faked 
transformations etc. The problem is, when we want to instantiate a new 
transformation writer sample, the script will fail because of the wrong 
namespace of Eigen::Vector3. The problem occurs on every artemis stack. 
The script-part looks like this:

################
# Faking the transformations
     trWriter = localPlanner.dynamic_transformations.writer()
     trSample = trWriter.new_sample()
     trSample.targetFrame = 'odometry'
     trSample.sourceFrame = 'body'
     trSample.time = Time.now()
     trSample.position = Eigen::Vector3.new(1,1,0)
     trSample.orientation = Eigen::Quaternion::Identity()

     trWriter.write(trSample)
#######################


Am 14.01.2015 um 19:34 schrieb Sylvain Joyeux:

doudou:
Could you give the exact error and backtrace ?

hemker:
Hey,

I was not able to get the same error now, because my group members 
changed a lot of code during the last week. Probably the ruby errors are 
not shown on stdout because other modules flush the output all the time; 
but if i change the spline.rb back to the original commit, the thread 
which alters the local planner's position terminates directly and 
nothing happens. Are there any ruby logfiles, generated from rock?

Am 15.01.2015 um 20:50 schrieb Sylvain Joyeux:

doudou:
I've done some tests and could not reproduce at all. You are talking about spline.rb, is that one of your file or the spline.rb from base/types ?

hemker:
The file is definetely located in sth. like Base/types/binding/ruby/lib/geometry/

---- Sylvain Joyeux schrieb ----

doudou:
Then I am even more confused. The code snippet you've shown before is definitely *your* code. Now you are talking about spline.rb which is within Rock.... What was the error exactly, and where was it located ?

hemker:
Ok,
We set up a little test script for testing the local and global planner located in planning/orogen/motion_planning_libraries/scripts. We have no real odometry and stuff, so we fake the new positions for the local planner. When the local planner finished planning, we set his new position to the last point of his planned trajectory. Then he continues planning till he reaches the global endpoint. For changing the position we use the code snippet I already sent you. Then the code fails because the wrong eigen ruby vector3 type. If I add the :: in the base types spline.rb everything is fine... 

---- Sylvain Joyeux schrieb ----

doudou:
I get it now ... I am looking into it

doudou:
Could you check if #25 fixes the problem ? It is a more generic fix.

hemker:
Yeah, it's doing it!


Am 16.01.2015 um 20:18 schrieb Sylvain Joyeux:

PR #25, opened 2015-01-16
ruby: define the Spline bindings outside Types (closes #24)

doudou:
The Types module is a *dynamic* module. Meaning that we never are
really sure about which are the constants that will be defined in
there, and therefore constant resolution is really fragile. In
consequence, manually writing code within this module *is* really
fragile.

For instance, the Types::Base::Geometry::Spline was referring to
Eigen::Vector3, which broke as soon as one type is loaded that
makes typelib create Types::Eigen (in which Vector3 is obviously
absent).

This commit moves the spline code into a SISL module, and only aliases
it (for backward compatibility reasons) in base/geometry/spline. It
also adds tests for the typelib-to-ruby conversions


doudou:
tested by @hemker in #24. In general, we must avoid writing code under the Types namespace at all from now on.

PR #27, opened 2015-01-20
Update logging

jakobs:


doudou:
ping @jakobs

2maz:
I just merged the uncritial bits / fixes (commit: 028d446 and  40b84f2).
The remaining point of discussions: how the logger should behave in Release mode and others and how to control the behaviour?

doudou:
> I just merged the uncritial bits / fixes (commit: 028d446 and 40b84f2).

@jakobs: could you rebase ?

jakobs:
Just rebased. Still need to decide where the decision to build in the logging comes from. In my opinion this should be decided in the cmake, which will then set the log level explicitly.

goldhoorn:
Do we need to make a public decision here?
Setting on all projects a debug level is in autoproj quite easy:

pkg.define("BASE_LOG_PRIORITY",5) for example....

doudou:
@goldhoorn: :+1:

goldhoorn:
In rock we have no default way of building a release (yet) if someone intorduce this (e.g. setting stable to all packages) then he could add this too. For the time beeing i merge this...

PR #28, opened 2015-02-05
Multiple Fixes / Additions to Spline, Angle, Pose2D

jmachowinski:


base/Pose.hpp
```diff
@@ -255,7 +256,24 @@ namespace base
         Pose2D( const Pose &p)
             : position(Vector2d(p.position.x(), p.position.y())), orientation(p.getYaw()) {}
 
+        bool isApprox(const Pose2D &other, double distPecision, double anglePrecision) const
+        {
+            return ((other.position - position).norm() < distPecision) && 
+                   (fabs((Angle::fromRad(other.orientation) - Angle::fromRad(orientation)).getRad()) < anglePrecision);
```
doudou: Would make a lot more sense to use the isApprox method from above.

src/Spline.cpp
```diff
@@ -216,7 +231,7 @@ double SplineBase::getCurvatureMax()
     if (has_curvature_max)
         return curvature_max;
 
-    double const delPara = getUnitParameter() * geometric_resolution;
+    double const delPara = geometric_resolution;
```
doudou: This change makes no sense. The search is in parameter space, geometric_resolution in geometrical space. If you would like to search in geometrical space, you would need to do a binary search to "advance" by geometric_resolution each point.

I seem to remember there is a way to (1) get the derivative curve of the spline and (2) get the max of a curve. That would also be a much better option.

src/Spline.cpp
```diff
@@ -200,10 +200,25 @@ double base::geometry::SplineBase::getCurveLength(double relative_resolution) co
     return length;
 }
 
-double SplineBase::getUnitParameter()
```
doudou: Why do you remove it ?

doudou: Just checked the commit log, sorry. There is kind of a relationship in "regular" curves (which is what SISL tries to produce).

The issue you're facing here is that by removing it you make the situation worse by using geometric_resolution as a parameter iteration, which is even most wrong.

base/Pose.hpp
jmachowinski: agreed, updated

src/Spline.cpp
jmachowinski: It was a constant factor before and still is one. Which is wrong... One could multiply it with 0.1 to make it finer grained or something. Still the result may be unexpected.

src/Spline.cpp
jmachowinski: It was a constant factor before and still is one. Which is wrong... One could multiply it with 0.1 to make it finer grained or something. Still the result may be unexpected. 

Agreed, the method is broken (and was before) but this patchset does not aim at fixing it.

src/Spline.cpp
jmachowinski: Agreed, feel free to add a patch

src/Spline.cpp
doudou: If the new code is as broken as the existing code, then leave the existing code. At least, it does not change the class behaviour.

jmachowinski: I added a patch to restore the original behaviour of getCuratureMax()

base/Angle.hpp
```diff
@@ -120,7 +120,7 @@ class Angle
      */
     bool inline isApprox( Angle other, double prec = 1e-5 ) const
     {
-	return fabs( other.rad - rad ) < prec;
+        return Angle(fabs( other.rad - rad )).getRad() < prec;
```
doudou: Angle(fabs(other-this)).getRad() can be negative if e.g. other-this is in [PI,2*PI]+. You want fabs(Angle(other.rad - rad).getRad()) (and you definitely will want to add a test for this particular corner case ...)

jmachowinski: Fixed

base/Pose.hpp
```diff
@@ -255,7 +256,24 @@ namespace base
...
     };
+    
+    inline std::ostream& operator << (std::ostream& io, base::Pose2D const& pose)
+    {
+        
+        io << "Position "
+           << pose.position.transpose()
+           << " Orientation (Theta) " 
+           << pose.orientation ;
+;
```
jakobs: artefact?

jakobs:
can you maybe split the spline stuff off into a separate PR?

doudou:
> can you maybe split the spline stuff off into a separate PR?

Would have been great, but I have no objection merging it as-is now.

base/Pose.hpp
jmachowinski: ?? 
This is a default stream operator for Pose2D. Pretty useful.

jmachowinski:
All Spline related changes are already in their own patches, why split it into several pull requests ?

base/Pose.hpp
jakobs: no, the ; at the beginning of the line.

src/Spline.cpp
planthaber: This removal brakes planning/corridor_planner.

I didn't get what is the alternative here.

@D-Alex 

>/home/planthaber/dfki/rock/entern/planning/corridor_planner/src/main.cc: In function 'void markCurve(base::geometry::Spline<3>&, int, std::vector<RGBColor>&, RGBColor)':
/home/planthaber/dfki/rock/entern/planning/corridor_planner/src/main.cc:51:27: error: 'class base::geometry::Spline<3>' has no member named 'getUnitParameter'
/home/planthaber/dfki/rock/entern/planning/corridor_planner/src/main.cc: In function 'void outputPlan(int, int, const string&, const std::vector<unsigned char>&, corridor_planner::Plan&)':
/home/planthaber/dfki/rock/entern/planning/corridor_planner/src/main.cc:165:43: error: 'class base::geometry::Spline<3>' has no member named 'getUnitParameter'
/home/planthaber/dfki/rock/entern/planning/corridor_planner/src/main.cc:182:47: error: 'class base::geometry::Spline<3>' has no member named 'getUnitParameter'
ICECC[7551] 10:21:09: Compiled on 192.168.128.14
src/CMakeFiles/plan_corridors.dir/build.make:54: recipe for target 'src/CMakeFiles/plan_corridors.dir/main.cc.o' failed

PR #29, opened 2015-02-06
Fixed color ordering of pointcloudes

goldhoorn:
This patch fixed VizkitPointCloudeVisualization


goldhoorn:
pong ?

goldhoorn:
bit importand this fixed a memory out-of-array-acces (len + 1 access)

doudou:
I really can't comment on this. Can anybody that has experience with these viz3d plugins have a look ? @jakobs ? @D-Alex ?

goldhoorn:
It's really only a minor fix.

The problem was that i was increased before the first access.
therefore pointCloud.colors[i][0] index 0 was never used. The last-used one was size()+1.
Really stupid misstake

The index of colors should equal the index of the iterator access but i don't want to change to a integer based iterator for the points-list (which would be clearer then).

jakobs:
I merge it, but why not do a proper fix like you suggest?

PR #32, opened 2015-02-11
add support to move a loaded model w.r.t the reference RBS

doudou:
When using loadModel, it is possible that the rigidbodystate being visualized is not the same than the model's reference frame. Instead of requiring a modification to the model, allow to specify the translation and rotation between the two.


doudou:
Superseded by #38

PR #35, opened 2015-02-12
move TransformWithUncertainty from envire to base-types

saarnold:
As part of the current restructuring of envire we identified that the type TransformWithUncertainty shall become a base type. It provides an Affine3d transformation with a corresponding full 6x6 covariance matrix and provides the composition methods handling both.

I made it a sample, since I think it needs also a timestamp. Ideally this type could be used in the future as alternative to the RigidBodyState, or even as a replacement, in the transformer component.

Since the Eigen::Affine3d stores its matrix as private member, I wrote an opaque conversion to Matrix4d.

Related PR: https://github.com/rock-core/base-orogen-types/pull/11

### DIFF
--- a/base/TransformWithUncertainty.hpp
+++ b/base/TransformWithUncertainty.hpp
@@ -1,0 +1,372 @@
+#ifndef __BASE_TRANSFORM_WITH_UNCERTAINTY_HPP__
+#define __BASE_TRANSFORM_WITH_UNCERTAINTY_HPP__
+
+#include <Eigen/Core>
+#include <Eigen/LU>
+#include <Eigen/Geometry>
+#include <Eigen/SVD>
+
+#include <base/samples/RigidBodyState.hpp>
+#include <base/Eigen.hpp>
+#include <base/Pose.hpp>
+
+namespace base {
+
+    /** 
+     * Class which represents a 3D Transform with associated uncertainty information.
+     *
+     * The uncertainty is represented as a 6x6 matrix, which is the covariance
+     * matrix of the [r t] representation of the error. Here r is the rotational
+     * part expressed as a scaled axis of rotation, and t the translational
+     * component.
+     *
+     * The uncertainty information is optional. The hasValidUncertainty() method can
+     * be used to see if uncertainty information is associated with the class.
+     */
+    class TransformWithUncertainty
+    {
+    public:
+	typedef base::Affine3d Transform;
+	typedef base::Matrix6d Covariance;
+
+	/** The transformation is represented as a 4x4 homogenous matrix. Both
+	 * rotation and translation in 3D are represented.
+	 */
+	Transform trans;
+	
+	/** The uncertainty is represented as a 6x6 matrix, which is the covariance
+	 * matrix of the [r t] representation of the error. Here r is the rotational
+	 * part expressed as a scaled axis of rotation, and t the translational
+	 * component.
+	 */
+	Covariance cov;
+
+	explicit TransformWithUncertainty( const Transform& trans = Transform::Identity() ) : 
+		trans( trans ) { invalidateUncertainty(); };
+
+	TransformWithUncertainty( const Transform& trans, const Covariance& cov ) :
+		trans( trans ), cov( cov ) {};
+
+	explicit TransformWithUncertainty( const base::samples::RigidBodyState& rbs )
+	{
+	    this->operator=( rbs );
+	};
+
+	static TransformWithUncertainty Identity()
+	{
+	    return TransformWithUncertainty( Transform::Identity() );
+	};
+	
+	
+	/** Default std::cout function
+	*/
+	friend std::ostream & operator<<(std::ostream &out, const TransformWithUncertainty& trans);
+
+	/** performs a composition of this transform with the transform given.
+	 * The result is another transform with result = this * trans
+	 */
+	TransformWithUncertainty composition( const TransformWithUncertainty& trans ) const
+	{
+	    return this->operator*( trans );
+	};
+
+	/** performs an inverse composition of two transformations.
+	 * The result is such that result * trans = this. Note that this is different from
+	 * calling result = this * inv(trans), in the way the uncertainties are handled.
+	 */
+	TransformWithUncertainty compositionInv( const TransformWithUncertainty& trans ) const
+	{
+	    const TransformWithUncertainty &tf(*this);
+	    const TransformWithUncertainty &t1(trans);
+	    Eigen::Affine3d t2 = tf.getTransform() * t1.getTransform().inverse( Eigen::Isometry );
+
+	    // short path if there is no uncertainty 
+	    if( !t1.hasValidUncertainty() && !tf.hasValidUncertainty() )
+		return TransformWithUncertainty( t2 );
+
+	    // convert the rotations of the respective transforms into quaternions
+	    // in order to inverse the covariances, we need to get both the t1 and t2 transformations
+	    // based on the composition tf = t2 * t1
+	    Eigen::Quaterniond 
+		q1( t1.getTransform().linear() ),
+		q2( t2.linear() );
+	    Eigen::Quaterniond q( q2 * q1 );
+
+	    // initialize resulting covariance
+	    Eigen::Matrix<double,6,6> cov = Eigen::Matrix<double,6,6>::Zero();
+
+	    Eigen::Matrix<double,6,6> J1;
+	    J1 << dr2r1_by_r1(q, q1, q2), Eigen::Matrix3d::Zero(),
+	    Eigen::Matrix3d::Zero(), t2.linear();
+
+	    Eigen::Matrix<double,6,6> J2;
+	    J2 << dr2r1_by_r2(q, q1, q2), Eigen::Matrix3d::Zero(),
+	    drx_by_dr(q2, t1.getTransform().translation()), Eigen::Matrix3d::Identity();
+
+	    cov = J2.inverse() * ( tf.getCovariance() - J1 * t1.getCovariance() * J1.transpose() ) * J2.transpose().inverse();
+
+	    // and return the resulting uncertainty transform
+	    return TransformWithUncertainty( 
+		    t2, cov );
+	};
+
+	/** Same as compositionInv, just that the result is such that trans * result = this.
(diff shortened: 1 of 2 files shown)


doudou:
Unless there's a bigger long-term plan to phase out RBS, I have a strong feeling against this. It doubes up with RigidBodyState

doudou:
Sorry, I actually did not read the description. /me bangs his head on the ground.

However, my comment stands. There's no way to convert from TransformWithUncertainty to RBS or the other way around, so by introducing this as a base type and (event worse) a base samples, you open the door to forking the dataflow between components using RBS and components using TransformWithUncertainty.

Which leads me to:
 - what would be the con of using RBS to represent TransformWithUncertainty on the wire
 - if you do need this type in base/types, I would keep it in base:: (not samples::) to make clear that it should not be used as a replacement to RBS.

saarnold:
> what would be the con of using RBS to represent TransformWithUncertainty on the wire

Mainly that one would lose information about the error. The error for the position and orientation is threaded uncoupled in the RBS case.

> if you do need this type in base/types, I would keep it in base:: (not samples::) to make clear that it should not be used as a replacement to RBS. 

I actualy also don't see it as a sample type, but I really would like to have it a timestamp. So I would put it in base but with a timestamp.

jakobs:
I actually think that TransformWithUncertainty (TWU) would be the more appropriate type to represent transformations. RBS with its velocity component and its broken covariance should (in the ideal world) be replaced. What about deriving RBS from TWU? It is a subset... The only thing that would not work is  cov_position and and cov_orientation. They are not used so much though, and could be easily updated to a function with access the covariance as a submatrix.

jakobs:
ah... sorry, actually TWU uses Isometry3 transform and RBS uses the more compact Quat+Vector to represent the transform. I think this could be easily changed in TWU, though.

saarnold:
> I actually think that TransformWithUncertainty (TWU) would be the more appropriate type to represent transformations.

I agree. 
And derivating RBS from TWU would be possible. But I would avoid it, because I don't like the idea of touching the RBS and a derivated RBS would be overloaded with methods and information that is most likely not required anymore.
IMO the RBS is the type that should be used in the case of holding single measurements, in which only a subset could be defined. The TWU should be used to represent transformation frames and therefore always provide a full transformation matrix.

> actually TWU uses Isometry3 transform and RBS uses the more compact Quat+Vector to represent the transform. I think this could be easily changed in TWU, though.

Using an Isometry3 transform should ensure that not only a subset of the transformation is actually defined and saves computational cost.

doudou:
While I would greatly appreciate the deprecation and removal of RBS, this is something that obviously cannot be done overnight. In other words, there should be a migration plan (missing rock core feature and implementation plan for them, announcements, clear guidelines on when to use which type and how to update existing components, required documentation updates). RBS is probably *the* most used type on orogen interfaces. Just dropping it in base/orogen/types is the first stone towards a really messy situation.

jhidalgocarrio:
Here my opinion on the discussion:
TWU is not a replacement of RBS but it needs to be in base/types, since the type time+transform+uncertainty is missing in the core.
TWU needs to have a timestamp. Therefore, I think base/types/base/samples is a good location.
Quaternion/AngleAxis representation should be better than Matrix transformation, because TWU is meant for Affine3d.
In a long term TWU should be the data type for the transformer and RBS for the state vector out information(body state). Obviously it cannot change overnight I think we could create a RBS <->TWU (TWU already have one) for the transformer to work with both types.
TWU will play nice with the transformer. It needs to have two std::string for source_frame and target_frame.
I agree we need a plan, but this already sounds like a good starting point!

doudou:
> In a long term TWU should be the data type for the transformer and RBS for the state vector out information(body state). Obviously it cannot change overnight I think we could create a RBS <->TWU (TWU already have one) for the transformer to work with both types. 

It's not as simple. Components will need to have both types as I/O or some will work only with TWU and some only with RBS

This is *not* a good starting point. As soon as TWU is in base, the obvious question everyone will have (including myself) is: "I have component X, which type should I use ?". Making sure that we have the answer *before* the type appears is IMO a must have. This kind of core changes cannot be done like this.

Moreover, if the overall goal is to use TWU as input to the transformer, I would use the opportunity to revisit how the transformer exchanges data over ports. I personally really have a problem with the source_frame/target_frame thing.

I guess, for me, the real question is: is there an absolute urgency to get it in base ? Again, I would welcome the change, but *if it gets properly planned and executed*.

jhidalgocarrio:
> This is not a good starting point. As soon as TWU is in base, the obvious question everyone will have (including myself) is: "I have component X, which type should I use ?". Making sure that we have the answer before the type appears is IMO a must have. This kind of core changes cannot be done like this.

I think the answer is simple. Does the component have state vector information (meaning velocity, angular velocity, etc..)? Yes, then port out a RBS. No, port out a TWU.

> Moreover, if the overall goal is to use TWU as input to the transformer, I would use the opportunity to revisit how the transformer exchanges data over ports. I personally really have a problem with the source_frame/target_frame thing.

I completely agree with you. source/target_frame is totally sensor processing information and does not make sense for representing kinematics chains/trees. It can be an opportunity to document and revisit this point 

Also it can be another opportunity to think about base/Pose.hpp. I see more overlapping with this type than with RBS itself.

> I guess, for me, the real question is: is there an absolute urgency to get it in base ? Again, I would welcome the change, but if it gets properly planned and executed.

Plan:
What about "not accepting" this pull request. However, it will be another one in the near future(we need some urgency) with the comments we already got:
1. Move to a more compact representation based on 6d axis+vector instead of matrix transform.
2. Add frame names. Probably with the same name source/target but with a proper meaning to allow kinematics chains/tree (useful for envire and manipulators)
3. What about this one? having TWU we cloud get rid of base/Pose and then TWU can be use by RBS to get the position and orientation. Methods setPose() and getPose() of RBS. It will make things  easier in the future.

One the new pull request is accepted:
We could start to adapt the transformer to work with TWU and since point 3 is already in RBS it will be simpler since RBS getPose() could return a TWU and therefore being the method interface between RBS and transformer (using TWU as data-type).

jakobs:
Fully agreed. So what should we do with this? 
Allow the merge now, and come up with a transition plan later?

doudou:
Hey. What about you create a page on the wiki and start laying the migration/integration plan ? Including relationships with the current data types.

Do remember that you just cannot connect two ports that don't have the same time. Introducing TWU as the transform type (as opposed to RBS) 

Also: what about having a base::samples::Joints-like type instead of the current RBS-derived system ? Transform frames would then be optional, and stored with neither timestamp nor names, with additional data stored alongside them (but optional).

jhidalgocarrio:
> @jakobs Allow the merge now, and come up with a transition plan later?
I guess wait for a new pull request, in order to have the TWU having 6d (axis vector) and RBS using the TWU in getPose and setPose methods. But that one would have more changes in base types than this current one.

> @doudou  Also: what about having a base::samples::Joints-like type instead of the current RBS-derived system ? Transform frames would then be optional, and stored with neither timestamp nor names, with additional data stored alongside them (but optional).

I don't fully understand this part, but I don't see as bad idea to have frame names as mandatory, because a transformation is always between two frames coordinates..
There is a wiki: https://github.com/rock-core/base-types/wiki/TranformWithUncertainty-in-base-types

saarnold:
I sadly did have time to join the videoconference today, but I agree with the decision.

The type is now in base::TransformWithUncertainty and doesn't have a timestamp.

jakobs:
The question on the compactness still stands. Although, to be fair, the covariance matrix has a much bigger impact on compactness. In the covariance we have 36 floats, of which 15 values are unused. Even when just storing the triangular matrix, this wouldn't be the most compact form. Do get a really compact version of the covariance matrix we would have to do an SVD. I think this would result in 18 values as the most compact form. The Transform part  as it is now as a 4x4 matrix takes 16 values. The most compact form would be 6. 

Most compact TWU would be 24 values. 
Current TWU is 52 values. 

In my opinion, just compacting the transform without compacting the covariance makes no sense. We should just leave it like it is now, but make sure we use Eigen::Isometry instead of Affine. Which I think we should also change in the base typedef.

doudou:
Ping @jakobs: what's left to do ? Change the typedef to Isometry and that's it ?

jhidalgocarrio:
Redirecting ping to @jhidalgocarrio...
Yes, I will take it. However, I am confuse with the wiki comment:
"base::TransformWithUncertainty", to me it should be base::samples::TransformWithUncertainty. I see TWU as an enhancement of base::Transform. TWU enhances base::Transforms by 1) Timestamp and 2) Uncertainty.

doudou:
Basically sascha went for the easy-to-merge solution: push TransformWithUncertainty as a non-sample type (and not-used-over-orogen type) to get it in base/types and have the time to get the orogen side of things ironed out. 

TWU enhances base::Transforms with uncertainty. Adding frames and timestamps is the separate matter (as you note yourself) of replacing RBS by TWU. As it is riddled with problems, let's get base::TWU merged right now and get the samples:: part ironed out.

jhidalgocarrio:
Alright. Therefore, don't wait to make TWU Isometry since base::Transform is not Isometry either. Then +1 to merge right now and close this issue. We would make further improvements  base on the wiki comments.

doudou:
Sounds good to me. I'd like to have an ack from the others though ...

saarnold:
Fine for me.
@jhidalgocarrio Didn't you favor the SVD of the covariance matrix?

jhidalgocarrio:
@doudou before to merge this. Can we discuss the topic tomorrow during the rock teleconf?

doudou:
> @doudou before to merge this. Can we discuss the topic tomorrow during the rock teleconf?

Sure.

saarnold:
What is the state here? Can it be merged? :)

jhidalgocarrio:
I am working in a forked branch not just to include TW as a new type but integrating it in base with a BodyState. I need to see how much it would affect base/types
You can have a look at https://github.com/envire
IMO we could remove this PR without merging it. Then, we can work and test the forked branch in envire until we get something that makes sense to pull-request to rock-core.

doudou:
Is that supserseded by #51 ?

saarnold:
Yes, it is. I'll close this PR.

PR #37, opened 2015-02-15
Changed wrong command

MalgosiaG:


goldhoorn:
This is a critical bug, it breaks all rock-build please merge asap
@doudou @jakobs @2maz

PR #41, opened 2015-03-03
Fixed findOneClosestPoint method.

jmachowinski:
The method was returning wrong closest points,
as closeness was computed by distance in param space.
Also it defaulted to return a point closest to guess,
what gave wrong results in some cases : E.g:
ClosestPoints returned params [0, 0.75], were 0.75 was
exactly _pt. As guess was 0, 0 was returned.

### DIFF
--- a/src/Spline.hpp
+++ b/src/Spline.hpp
@@ -259,6 +259,10 @@ namespace geometry {
                 std::vector<double>& points,
                 std::vector< std::pair<double, double> >& segments, double _geores) const;
 
+        /**
+         * Warning, do not use this method, it is broken and returns the wrong result
+         * 
+         * */
         double findOneClosestPoint(double const* _pt, double _guess, double _geores) const;
         void findClosestPoints(double const* ref_point,
                 std::vector<double>& _result_points,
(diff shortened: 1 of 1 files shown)


doudou:
> The method was returning wrong closest points, as closeness was computed by distance in param space.

The method was behaving as expected. You might want to change the behaviour if the new one is truly better, but you should at least argue about why it is better.

SISL returns the place of the curve (segments and points) that are within _geores of the closest point. Then, the value closest to the guess was returned.

While I do understand what you try to achieve, how do you deal with segments ? The "truly closest" point might be somewhere on a curve. You end up with inconsistent behaviour (a problem the original implementation did not have) as you will in this case return one of the two boundaries while for the other cases you return the "true closest".

IMO, just think very hard about which geores makes sense for your application and stick with the consistent behaviour.

jmachowinski:
> The method was behaving as expected. 

The documentation of the method said in the first line
'Returns a single closest point to _pt' 
which it did not. It returned the closest point to the guess 
(I known as written below). 

The old implementation returns wrong values for me, even
with arbitrarily low geores. The new implementation always
returned the expected values. Either start or end of the subcurve 
seem to be alway within geores of the searched point.

The old implementation showed a 'edgy' behavior. The start param
of the subcurve seem to get increased in an internal step width. 
This lead to the behavior that the closest point jumped.

jmachowinski:
So this is ready for merge ?

doudou:
Forgot that one. If findClosesPoints is the problem, why don't you filter findClosestPoints. You've found a bug, and let it there for others to hit ?

jmachowinski:
findClosestPoints works as expected. The problem in findOneClostestPoint is the check being based on the param values, and not an the real distance.

jmachowinski:
Ping

PR #42, opened 2015-03-17
Trajectory visualization: use a method instead of a block as callback

saarnold:
Using a block here currently leads to the problem that only the latest block is called with samples of both output ports, when attaching a second trajectory vizkit plugin to another output port.

There is apparently a problem of how the block callbacks are handled in vizkit.


doudou:
Could you try rock-core/gui-vizkit#23 ? It fixed the underlying problem for me.

doudou:
Fixed in vizkit itself, https://github.com/rock-core/gui-vizkit/pull/23

PR #45, opened 2015-05-06
Fixed datatypes for spline handling

shoppel:


goldhoorn:
Might be a workaround here but occurred nowhere else sometimes the ruby datatype binding get's typelib types instead of atomic types.

Maybe you know why @D-Alex (to take load off sylvain)

goldhoorn:
<b> needs to be patched ro rock-rc if accepted </b>

D-Alex:
I would no merge this code. We should try to find out why the original code is not working. Can you provide a minimal example how to reproduce the error, add an issue to (I guess it would be typelib) and reference this pull request

goldhoorn:
I just discovert that error #8 

i pushed a simple testcase (which is i assume related to this) install this version of the message-produver and run it and show it in rock-display:

https://github.com/goldhoorn/tutorials-orogen-message_producer

It would be great if someone else could do this i'm quite busy with the release in general
@D-Alex or @marvin2k (marvin you are in typekit therefore i added you)

goldhoorn:
fixed by https://github.com/rock-core/gui-vizkit/pull/34

PR #46, opened 2015-05-06
viz: simplify sonar beam visualization code

doudou:
Instead of doing its own management of the sonar stream, it now uses the
keepOldData features to achieve the same result as the existing code, but
a lot simpler (and easy to change at runtime). The beams are now
accurately represented (i.e. are plus/minus a half-width around the
center bearing). Finally, bearing=0 means beam along X.

### DIFF
--- a/viz/SonarBeamVisualization.cpp
+++ b/viz/SonarBeamVisualization.cpp
@@ -9,6 +9,8 @@ SonarBeamVisualization::SonarBeamVisualization()
 {
     bodyState.invalidate();
     newSonarScan = false;
+    setKeepOldData(true);
+    setMaxOldData(100);
 }
 
 osg::ref_ptr< osg::Node > SonarBeamVisualization::createMainNode()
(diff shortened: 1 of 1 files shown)


D-Alex:
:+1: 

I just went through the code and added some comments while reading. But even without the changes the code is a good improvements compared to the original one.

doudou:
Hey. Next time, make sure that you comment in the "FIles changes" section of the PR, this makes looking at them really a lot easier, and makes sure that we don't lose any comment if the submitter rebases or amends commits.

D-Alex:
sorry, I did not know there is a different.

doudou:
Other than that, I've updated the code according to your comments. Can you have a look and merge if you're fine with the result ?

D-Alex:
I cannot see any changes? - have you pushed to a different branch?

goldhoorn:
@doudou did you forgot this?

doudou:
Nope... but did not have time to fix / update / ...

PR #48, opened 2015-06-08
fix minor Typos in base/types

leifole:


goldhoorn:
:+1: for merge, will apply on 10.06.2015 to avoid to much recompilation.

PR #51, opened 2015-06-09
New body state

jhidalgocarrio:
This is the aim of a really needed base type for everyone doing slam in rock.

This PR will include 3 new types:
1. TransformWithCovariance in [AngleAxisd Translation] 6D position vector with covariance for representing uncertainty
2. TwistWithCovariance in [rotation translation] 6D velocity vector with covariance.
3. BodyState for propagarion of the body pose and velocity with the full covariance. No frame names.

The propagation of the uncertainty follows the same equations than TransformWithUncertainty from the envire library and it aims to replace it. This PR does not aim to replace the existing RigidBodyState (RBS) since BodyState does not have frame names. It aims to live with RBS for the short-middle time.

With BodyState it is possible to perform composition operations propagating the uncertainty.
body2 = body2 * body1


jhidalgocarrio:
The current implementation uses screw theory notation (first rotation and then translation) for Pose and Twist. Therefore the information in the covariance matrix is also in that order. This convention was also used at the previous envire TransformWithUncertainty but it should be notice here that this is the apposite from ROS REP 103 (x, y, z, rotation about X axis, rotation about Y axis, rotation about Z axis).

I would like to have an agreement on that aspect.

jhidalgocarrio:
Comments on this?
ping....

doudou:
Hi Javier. I am very sorry that you did not get any feedback on this. Thanks a lot for this work ! I will try very very hard to take the time to look at it more closely.

Is there a drawback for sticking to ROS compatibility here ? Would that be a lot of work ? (i.e. do you depend in your code on the representation, or do you convert to Eigen types anyways ?)

jhidalgocarrio:
Great, I would like to have some feedback.

Is there a drawback for sticking to ROS compatibility here ?
No. Actually after using this PR during the last month I feel more to switch  ROS REP 103 convention.
Would that be a lot of work ? 
No. just changing order of internal covariance storage in TransformWithCovariance(TWC). The API stays and we uses Eigen types anyway.

I also suggest to change orientation from AngleAxis to Quaternion in the TWC class. It will be closer to ROS REP 103 and the RigidBodyState class. I originally chose AngleAxis to have 3 elements but internally Eigen uses 4 elements because it is not scaled-axis, so... in order to choose a 4-element representation a Quaternion will better fit to the modified TWC in base/types (for historical reasons)

doudou:
About the change to quaternion: honestly, that's your choice, I am mostly looking into the potential integration/architecture problems - not very fit to comment on the theory. Being compatible to ROS conventions definitely sounds good if there's no major drawback.

jhidalgocarrio:
I will change AngleAxis to Quaternion and change the covariance storage to ROS REP 103 order instead of the current order coming from envire.

@doudou When I have this done, I would like to merge this PR since it is quite urgent. Does it sound good to you?

base/Pose.hpp
```diff
@@ -37,35 +39,89 @@ namespace base
...
 
+    static double getRoll(const base::AngleAxisd& orientation)
+    {
+        return base::getEuler(orientation)[2];
+    }
 
     static inline base::Orientation removeYaw(const base::Orientation& orientation)
     {
-	return Eigen::AngleAxisd( -getYaw(orientation), Eigen::Vector3d::UnitZ()) * orientation;
+	    return Eigen::AngleAxisd( -getYaw(orientation), Eigen::Vector3d::UnitZ()) * orientation;
```
doudou: You indent with tabs and have your editor setup with tab=4 spaces. Fucks up indentation on e.g. github and anyone that has tab == 8 spaces (which is the default for most editors). Please stick to 4-space indent, no tabs.

base/samples/BodyState.hpp
```diff
@@ -0,0 +1,337 @@
...
+        {
+            return this->velocity.getAngularVelocityCov();
+        }
+
+        inline void cov_angular_velocity(const base::Matrix3d& cov)
+        {
+            return this->velocity.setAngularVelocityCov(cov);
+        }
+
+        static BodyState unknown()
```
doudou: I would suggest naming them Invalid() and Unknown() (i.e. using uppercase) to match Eigen's Zero, Identity, ...

doudou:
@jhidalgocarrio: no further blockers from me (apart from the few comments I just added). Sorry for the long long delay ...

doudou:
(NOTE: you will have to rebase or merge manually, there are conflicts)

jhidalgocarrio:
Ready to merge with master (sorry for my delay)

goldhoorn:
I merged and tested on master wihth all packages -> no negative impact

PR #52, opened 2015-06-10
ruby: fix signed-ness of size-related methods in Eigenn bindings

goldhoorn:
The methods were returning a int instead a unsinged int. This fixes compiler warnings in base/orogen/std. The Eigen methods return unsigned ints already.


doudou:
How is that related to base/orogen/std ????

goldhoorn:
Sorry base/orogen/types (the tests give warning regarding comparism between signed and unsigned strucures)

doudou:
Ah ... right.

PR #53, opened 2015-06-11
remove base/geometry/spline

doudou:
It is incompatible with the new registry export mechanism in
Typelib's ruby bindings.


2maz:
As you say it is backward-incompatible. Is there no alternative to maintain the current behaviour or do you just intend to fix it later?

doudou:
No real alternative that I can see. The problem is that the new typelib behaviour really expects that noone would define modules under the Types namespace (which, by the way, was a bad idea in the first place)

goldhoorn:
ack for me

2maz:
ok then, merged

PR #59, opened 2015-08-14
samples/Sonar: define base::samples::Sonar

doudou:


base/samples/Sonar.hpp
```diff
@@ -0,0 +1,289 @@
...
+     *
+     * Bin values must be normalized, i.e. gain effects must be corrected and -
+     * as much as made possible by the sonar's specification - converted to a
+     * linear scale where 1.0 means that the signal received by the transducer
+     * is equal to the signal transmittd.
+     *
+     * The data is stored beam-first, that is the data for beam N starts at N *
+     * bin_count and ends at (N + 1) * bin_count
+     */
+    std::vector<float> bins;
```
goldhoorn: Did we finally decided to switch to floats?, otherwise i would recommend to stay to doubles for all base-types until we made a final-decision.

base/samples/Sonar.hpp
```diff
@@ -0,0 +1,289 @@
...
+        : speed_of_sound(1497)
+    {
+        init(time, beam_count, bin_count, init_to_unknown);
+    }
+
+    /** Initializes a Sonar structure to represent a single beam
+     */
+    static void fromSingleBeam(base::Time time, std::vector<float> const& bins, base::Angle bearing = base::Angle())
+    {
+        Sonar sample;(time, 1, 0, false);
```
goldhoorn: This this a typo, or some new c++ style i don't know?

base/samples/Sonar.hpp
```diff
@@ -0,0 +1,289 @@
...
+    void validate()
+    {
+        if (bin_count * beam_count != bins.size())
+            throw std::logic_error("the number of elements in 'bins' does not match the bin and beam counts");
+        if (!timestamps.empty() && timestamps.size() != beam_count)
+            throw std::logic_error("the number of elements in 'timestamps' does not match the beam count");
+        if (bearings.size() != beam_count)
+            throw std::logic_error("the number of elements in 'bearings' does not match the beam count");
+    }
+};
```
goldhoorn: Maybe it would make sense to create a constructor/Setter from the old SonarScan type to make backward-compatibility easier...

base/samples/Sonar.hpp
doudou: Definitely a typo ... weird that it compiled.

base/samples/Sonar.hpp
doudou: Sounds like a great idea.

base/samples/Sonar.hpp
doudou: My understanding was that we should use float unless there is a compelling reason to use a double. Actually, people started using floats and doubles around (I took the float from the DepthMap type).

base/samples/Sonar.hpp
doudou: It did not. I thought I tried compilation but somehow did not.

doudou:
Updated with backward-compatible code, and some missing fields.

base/samples/Sonar.hpp
saarnold: +1
I don't see a reason to use a higher resolution here.

doudou:
@goldhoorn, @saarnold: should I go ahead and merge?

base/samples/Sonar.hpp
```diff
@@ -0,0 +1,433 @@
...
+     * is equal to the signal transmittd.
+     *
+     * The data is stored beam-first, that is the data for beam N starts at N *
+     * bin_count and ends at (N + 1) * bin_count
+     */
+    std::vector<float> bins;
+
+    static float getSpeedOfSoundInWater() { return 1497.0; }
+
+    Sonar() : speed_of_sound(1497), bin_count(0), beam_count(0) {}
```
saarnold: Is there a reason why this constructor doesn't use the default speed of sound value provided by the static method?

doudou: Nope. Fixed, thanks !

base/samples/Sonar.hpp
```diff
@@ -0,0 +1,436 @@
...
+    {
+        return bin_duration * bin_idx;
+    }
+
+    /** The acquisition start of a beam
+     */
+    base::Time getBeamAcquisitionStartTime(int beam) const
+    {
+        if (timestamps.empty())
+            return timestamps.front();
```
saarnold: This should either throw an exception or return the time field.

base/samples/Sonar.hpp
```diff
@@ -0,0 +1,436 @@
...
+     * @arg interval the interval between two beams
+     *
+     */
+    void setRegularBeamBearings(base::Angle start, base::Angle interval)
+    {
+        base::Angle angle;
+        bearings.resize(beam_count);
+        for (uint32_t i = 0; i < beam_count; ++i, angle += interval)
+            bearings[i] = angle;
+    }
```
saarnold: The start angle is not set in this method.

saarnold:
The rest looks fine for me, but I would strongly recommend to add unit tests.
For the bin and beam index parameters one could use unsigned integers, this would prevent potential misusage.

doudou:
> For the bin and beam index parameters one could use unsigned integers, this would prevent potential misusage.

For the record, it only makes access at beams[0xFFFFFFFFF] instead of beams[-1] (which is more likely to crash, so that's good ... somehow)

> The rest looks fine for me, but I would strongly recommend to add unit tests.

Yes, came to that conclusion as well. I have to admit that I started doing it only because I saw *you* wrote tests for DepthMap ... Feeling a bit lonely in the unit test front.

doudou:
Done (acquisition start time), done (start angle), done (unsigned int) and done (unit tests).

And, as expected, fixed a bunch of issues that got detected by the tests (my favorite is https://github.com/rock-core/base-types/commit/2013847b17bc74e4ca072aaada65c40bf44881a1)

doudou:
Is it good to merge now ?

goldhoorn:
Nothing against it from my side

saarnold:
Neither from mine

PR #61, opened 2015-08-19
DepthMap: added an additional single time field

saarnold:
This time field is picked up by the stream aligner and transformer.
I noticed that this is missing for this type, when I was reviewing the new sonar type.

IMO the more cleaner option would be to inherit every type in samples from a timestamp interface, which implements a getTimestamp method the transformer could use.

### DIFF
--- a/base/samples/DepthMap.hpp
+++ b/base/samples/DepthMap.hpp
@@ -53,6 +53,13 @@ public:
 	UNIT_Y = 1,
 	UNIT_Z = 2
     };
+
+    /** Reference timestamp for the depth map sample.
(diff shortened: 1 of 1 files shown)


doudou:
Merged and then force-pushed to cancel it right away ... I hope nobody got hurt.

doudou:
Sorry, this mean you will have to re-open a new PR. The issue is that DepthMap is part of rock-15.05. At the very minimum, you will have to make sure people can update their log files (provide updated log migration scripts)

saarnold:
It is possible to update the log files anyway, the only issue is that the new time field will not be set.
I fixed it and will reopen the PR.

PR #62, opened 2015-08-26
DepthMap: added an additional single time field

saarnold:
This time field is picked up by the stream aligner and transformer.
I noticed that this is missing for this type, when I was reviewing the new sonar type.

IMO the more cleaner option would be to inherit every type in samples from a timestamp interface, which implements a getTimestamp method the transformer could use.

There is now also a conversion in rock-core/base-orogen-types#18 which sets the time field properly when converting from an existing DepthMap.


saarnold:
Since the type conversion update was already merged, this PR needs to be merged too.

PR #63, opened 2015-11-11
Optionally show the source frame of a RigidBodyState

chhtz:
This is useful when visualizing multiple frames in the same visualization.
To activate this, setTextSize to a positive value.
Warning: If the name of the sourceFrame changes, the visualization is not automatically updated. You can reset the size to do so.


goldhoorn:
Everything fine for me. Any remarks @D-Alex  otherwise I merge

goldhoorn:
Thanks for your contribution

PR #65, opened 2015-12-01
Visualize vector of RGBs

shoppel:
Allows to visualize vectors of RGBs, fixes some bugs (crash if a model could not be loaded), does some formatting and adapts the order of the methods within header and source file (reason why so much changes are shown).

Images can be load as textures again, but bump mapping is still not available (remove completely?).

### DIFF
--- a/viz/RigidBodyStateVisualization.cpp
+++ b/viz/RigidBodyStateVisualization.cpp
@@ -51,15 +51,22 @@ void RigidBodyStateVisualization::setColor(const Vec4d& color, Geode* geode)
 
 bool RigidBodyStateVisualization::isPositionDisplayForced() const
 { return forcePositionDisplay; }
+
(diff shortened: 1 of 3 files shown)


doudou:
This is getting ... too much IMO. Why not create a different plugin that uses this one as children ?

As for bump mapping, I'm using this, why do you want to remove it ?

shoppel:
I keep getting asked for the RGB vector visualization. Ok there are lots of changes (btw. mostly because of the formatting commit) but alltogether it is just a cleaned up, more stable version (e.g. no crashes if any path is wrong). Putting it in another viz plugin would just double the code.. The bump mapping did not work for me and I could not imagine any meaningful use-case, so I did not include it yet.

Maybe you can test und pull this nevertheless?

shoppel:
Ok and I took care that the header and the source file use the same method order, this created lots of changes as well (without changing anything).

PR #67, opened 2015-12-21
Add sonar conversion methods

romulogcerqueira:
New Sonar Datatype -> Legacy Sonar Datatypes


romulogcerqueira:
Closing here because it is the same as https://github.com/rock-core/base-types/pull/74

PR #68, opened 2016-01-27
fixed missing linker option for sisl in pkgconfig

jmachowinski:


doudou:
Unless there are direct calls to the SISL API from the Spline.hpp header (and last I looked there are none), having dependencies linking explicitely to SISL should not be needed.

jmachowinski:
You're right, the bug was in a sublibrary.

PR #72, opened 2016-03-10
Minor improvements in logging

marvin2k:
nothing special, just strolling by...


marvin2k:
I was temporarily confused, this needs relinking of all dynamic libraries... Handled by `amake smth`. Reopening ;-)

PR #74, opened 2016-03-22
Sonar datatype conversions

romulogcerqueira:
Conversions from the new sonar datatype (Sonar) to the legacy ones (SonarBeam and SonarScan).

### DIFF
--- a/base/samples/Sonar.hpp
+++ b/base/samples/Sonar.hpp
@@ -436,6 +436,23 @@ BASE_TYPES_DEPRECATED_SUPPRESS_START
             bins[i] = static_cast<float>(old.beam[i] * 1.0 / 255) * gain;
         pushBeam(bins, old.bearing);
     }
+
+    base::samples::SonarBeam convert2SonarBeam() {
+        base::samples::SonarBeam sonar_beam;
+        sonar_beam.time = time;
+        sonar_beam.speed_of_sound = speed_of_sound;
+        sonar_beam.beamwidth_horizontal = beam_width.rad;
+        sonar_beam.beamwidth_vertical = beam_height.rad;
+        sonar_beam.bearing = bearings[0];
+        sonar_beam.sampling_interval = 0;
+
+        std::vector<float> temp(bins.begin(), bins.end());
+        std::transform(temp.begin(), temp.end(), temp.begin(), std::bind2nd(std::multiplies<float>(), 255));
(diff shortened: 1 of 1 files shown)


base/samples/Sonar.hpp
```diff
@@ -436,6 +436,44 @@ BASE_TYPES_DEPRECATED_SUPPRESS_START
             bins[i] = static_cast<float>(old.beam[i] * 1.0 / 255) * gain;
         pushBeam(bins, old.bearing);
     }
+
+    base::samples::SonarBeam convert2SonarBeam() {
```
doudou: a more common pattern is to simply use the 'to' prefix, toSonarBeam, toSonarScan

romulogcerqueira:
Fixed, @doudou!

base/samples/Sonar.hpp
```diff
@@ -436,6 +436,44 @@ BASE_TYPES_DEPRECATED_SUPPRESS_START
...
+        base::samples::SonarBeam sonar_beam;
+        sonar_beam.time = time;
+        sonar_beam.speed_of_sound = speed_of_sound;
+        sonar_beam.beamwidth_horizontal = beam_width.rad;
+        sonar_beam.beamwidth_vertical = beam_height.rad;
+        sonar_beam.bearing = bearings[0];
+        sonar_beam.sampling_interval = 0;
+
+        std::vector<float> temp(bins.begin(), bins.end());
+        std::transform(temp.begin(), temp.end(), temp.begin(), std::bind2nd(std::multiplies<float>(), 255));
```
doudou: This assumes that the sonar data is in between [0, 1], which it is not. I'm not sure what's the best option. The best result would be to do an auto-gain (i.e. to compute the max value in self and scale that one to 255)

romulogcerqueira: You are right, @doudou. If any of sonar data isn't in between [0,1], I could throw an exception. What do you think?

> The best result would be to do an auto-gain (i.e. to compute the max value in self and scale that one to 255)

This would be a normalization?

doudou: > If any of sonar data isn't in between [0,1], I could throw an exception

You can't do that, really. I think it makes a lot more sense to just normalize the data.

romulogcerqueira:
update code, @doudou

PR #76, opened 2016-05-19
Make it possible to change the default color of DepthMap visualization.

chhtz:
Furthermore, make it possible to combine colorization and remission visualization.
Also, fixed some typos.

### DIFF
--- a/viz/DepthMapVisualization.cpp
+++ b/viz/DepthMapVisualization.cpp
@@ -20,6 +20,7 @@ DepthMapVisualization::DepthMapVisualization() :
 {
     scan_orientation = Eigen::Quaterniond::Identity();
     scan_position.setZero();
+    default_feature_color = osg::Vec4f(1.0f, 1.0f, 1.0f, 1.0f);
 }
 
 DepthMapVisualization::~DepthMapVisualization()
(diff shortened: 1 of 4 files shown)


viz/DepthMapVisualization.cpp
```diff
@@ -98,10 +100,15 @@ void DepthMapVisualization::updateMainNode ( osg::Node* node )
     }
     else if(show_remission && !scan_sample.remissions.empty())
     {
-        for(unsigned i = 0; i < scan_sample.remissions.size(); i++)
+        for(unsigned i = 0; i < points.size(); i++)
         {
 	    if(scan_sample.isIndexValid(i))
```
doudou: If you change the index from remissions to points, you probably should validate that points.size() == remissions.size().

chhtz: Done.

viz/DepthMapVisualization.cpp
```diff
@@ -113,7 +124,7 @@ void DepthMapVisualization::updateMainNode ( osg::Node* node )
     }
     else
     {
-        colors->push_back(osg::Vec4(1.0,0,0.3,0.8));
+        colors->push_back(default_feature_color);
```
doudou: Wouldn't it make more sense to keep the default feature color to its current value ? Not sure if/what you discussed with the primary users of this visualization ...

chhtz: I changed the default to the old color. That change was honestly merely for my convenience.

PR #77, opened 2016-05-25
Added heading to Motioncommand2d

Cirromulus:
To support omnidirectional drives I suggest adding a heading value to Motion2D-command.
(As suggested in comment of previous committer)

### DIFF
--- a/base/commands/Motion2D.hpp
+++ b/base/commands/Motion2D.hpp
@@ -13,10 +13,12 @@ namespace base
     {
         double translation; //! translation value in m/s
         double rotation;    //! rotation value in rad/s. Positive is counter-clockwise
-        // That's actually enough information for a differential drive,
-        // for an omnidirectional drive we would add a heading/angle
-        // value.
+        double heading;     //! heading in rad. Positive is counter-clockwise
+        Motion2D():translation(0), rotation(0), heading(0){};
+        Motion2D(double translation, double rotation, double heading):translation(translation), rotation(rotation), heading(heading){};
     };
+    inline bool operator==(const Motion2D& lhs, const Motion2D& rhs){ return lhs.translation == rhs.translation && lhs.rotation == rhs.rotation && lhs.heading == rhs.heading;}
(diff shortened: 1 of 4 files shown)


doudou:
MotionCommand2D represent velocities, not positions. I don't see how the heading is relevant.

Cirromulus:
Heading is relevant for omnidirectional drives, e.g for driving sideways. Furthermore it enables the device doing circles facing in any other direction than the driving-vector as well.
Heading is an angle that rotates the driving vector.

Eg.: 
Sideways would be [translation = 1, rotation = 0, heading = 90°]
Circling around a point in front of Device facing towards point would be [translation = 1, rotation = 1, heading -90°]

doudou:
OK ... got it. It's basically the direction of the translation vector w.r.t. the X axis of the vehicle.

My only comment then would be to use a base::Angle.

Cirromulus:
Ok, i'll change it

Cirromulus:
Better?

base/commands/Motion2D.hpp
```diff
@@ -13,9 +15,11 @@ namespace base
     {
         double translation; //! translation value in m/s
         double rotation;    //! rotation value in rad/s. Positive is counter-clockwise
-        double heading;     //! heading in rad. Positive is counter-clockwise
-        Motion2D():translation(0), rotation(0), heading(0){};
-        Motion2D(double translation, double rotation, double heading):translation(translation), rotation(rotation), heading(heading){};
+        base::Angle heading;     //! heading in rad. Positive is counter-clockwise
+        
+        Motion2D():translation(0), rotation(0), heading(base::Angle::fromRad(0)){};
+        Motion2D(double translation, double rotation):translation(translation), rotation(rotation), heading(base::Angle::fromRad(0)){};
```
jmachowinski: Better use the default constructor. 
I initializes with base::unset<double>. In this case 
one can distinguish between not used and Zero..

Cirromulus:
In this Construct Zero is equivalent to unset, because an Angle of 0 would not shift the Vector, as this would be the case before it was introduced.

jmachowinski:
Ok

PR #78, opened 2016-05-26
viz/PluginLoader bugfixes

arneboe:
This PR contains a minor and a major bugfix.

**Minor**: The QtPluginVizkit claimed to provide GridVisualization but it didnt. Fixed by removing the entry from `pluginNames`.

**Major**: `QtPluginVizkit` is also the name of the plugin factory in Vizkit3d. Thus when loading libbase-viz and libvizkit3d-viz you would get name clashes. Those led to strange behavior in Vizkit3d. This was fixed by renaming `QtPluginVizkit` to `QtPluginVizkitBase`.

### DIFF
--- a/viz/PluginLoader.cpp
+++ b/viz/PluginLoader.cpp
@@ -11,11 +11,11 @@
 #include "DepthMapVisualization.hpp"
 
 namespace vizkit3d {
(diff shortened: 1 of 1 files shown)


jhidalgocarrio:
+1 working in my local test :-)

PR #83, opened 2016-06-27
Add methods to get the min/max acceptable angles in base/Angle.hpp

romulogcerqueira:


doudou:
would be great to add unit tests, pls

romulogcerqueira:
@doudou, updated code!

base/Angle.hpp
```diff
@@ -106,6 +107,20 @@ class Angle
         return result;
     }
 
+    /** Use this method to get the minimum acceptable angle
```
jmachowinski: The documentation is wrong. It is the maximum angle returned by getRad(). If you use fromRad, you are allowed to give any value, and it will be normalized to the range from min <-> max.

doudou: > The documentation is wrong

The documentation is indeed ambiguous. I would suggest either "Returns the minimum Angle" or "Returns the minimum angle possible after normalization"

(I guess that Max would need to be changed accordingly)

romulogcerqueira: I agreed with you, @jmachowinski and @doudou. 

I followed the second @doudou's suggestion: "Returns the minimum angle possible after normalization".

Code is already updated.

jmachowinski:
looks good from my side

PR #85, opened 2016-06-28
New transformation type

saarnold:
This type shall be used in the future to express transformations.
I know that replacing the RigidBodyState will be a very long (and hard) process and I don't plan to force anyone to use this type. The idea is to first integrate this type in the back-end of the transformer while keeping the interfaces to the RigidBodyState and also add interfaces for the new type. At that point users could at least use the new type inside of their tasks and libraries already.

In contrast to the RigidBodyState this type has a full pose covariance matrix, no velocity members and a hopefully more intuitive frame naming.

### DIFF
--- a/src/samples/PoseWithCovariance.cpp
+++ b/src/samples/PoseWithCovariance.cpp
@@ -1,0 +1,75 @@
+#include "PoseWithCovariance.hpp"
+
+namespace base { namespace samples {
+
+PoseWithCovariance::PoseWithCovariance()
+{
+
+}
+
+PoseWithCovariance::PoseWithCovariance(const TransformWithCovariance& transform) :
+        transform(transform.translation, transform.orientation, transform.cov)
+{
+
+}
+
+PoseWithCovariance::PoseWithCovariance(const RigidBodyState& rbs) : time(rbs.time),
+    frame_id(rbs.targetFrame), object_frame_id(rbs.sourceFrame), transform(rbs.position, rbs.orientation)
+{
+    transform.cov << rbs.cov_position, Eigen::Matrix3d::Zero(),
+                        Eigen::Matrix3d::Zero(), rbs.cov_orientation;
+}
+
+PoseWithCovariance PoseWithCovariance::operator*(const PoseWithCovariance& pose) const
+{
+    // compose new pose
+    PoseWithCovariance new_pose;
+    new_pose.time.microseconds = std::max(this->time.microseconds, pose.time.microseconds);
+    new_pose.object_frame_id = pose.object_frame_id;
+    new_pose.frame_id = this->frame_id;
+    new_pose.transform = this->transform * pose.transform;
+    return new_pose;
+}
+
+void PoseWithCovariance::setTransform(const TransformWithCovariance& transform)
+{
+    this->transform = transform;
+}
+
+void PoseWithCovariance::setTransform(const Eigen::Affine3d& transform)
+{
+    this->transform.setTransform(transform);
+}
+
+const TransformWithCovariance& PoseWithCovariance::getTransformWithCovariance() const
+{
+    return transform;
+}
+
+Eigen::Affine3d PoseWithCovariance::getTransform() const
+{
+    return transform.getTransform();
+}
+
+const TransformWithCovariance::Covariance& PoseWithCovariance::getCovariance() const
+{
+    return transform.getCovariance();
+}
+
+RigidBodyState PoseWithCovariance::toRigidBodyState() const
+{
+    base::samples::RigidBodyState rbs;
+    rbs.time = time;
+    rbs.targetFrame = frame_id;
+    rbs.sourceFrame = object_frame_id;
+    rbs.position = transform.translation;
+    rbs.orientation = transform.orientation;
+    if(transform.hasValidCovariance())
+    {
+        rbs.cov_position = transform.getTranslationCov();
+        rbs.cov_orientation = transform.getOrientationCov();
+    }
+    return rbs;
+}
+
+}}
--- a/src/samples/PoseWithCovariance.hpp
+++ b/src/samples/PoseWithCovariance.hpp
@@ -1,0 +1,80 @@
+#ifndef BASE_SAMPLES_POSE_WITH_COVARIANCE_HPP
+#define BASE_SAMPLES_POSE_WITH_COVARIANCE_HPP
+
+#include <string>
+#include <base/Time.hpp>
+#include <base/TransformWithCovariance.hpp>
+#include <base/samples/RigidBodyState.hpp>
+
+namespace base { namespace samples {
+
+/**
+ * Represents a 3D pose with uncertainty of the frame 'object_frame_id' in the frame 'frame_id'.
+ * A frame ID shall be globally unique and correspond to a coordinate system.
+ * Following the [Rock's conventions](http://rock.opendfki.de/wiki/WikiStart/Standards)
+ * coordinate systems are right handed with X forward, Y left and Z up.
+ *
+ * The pose \f$ ^{reference}_{object}T \f$ in this structure transforms a vector \f$ v_{object} \f$
+ * from the object frame to the reference frame:
+ * \f$ v_{reference} = ^{reference}_{object}T \cdot v_{object} \f$
+ * Example of a transformation chain:
+ * \f$ ^{world}_{sensor}T = ^{world}_{body}T \cdot ^{body}_{sensor}T \f$
+ */
+class PoseWithCovariance
+{
+public:
+    /** Reference timestamp of the pose sample */
+    base::Time time;
+
+    /** ID of the reference frame where this pose is expressed in */
+    std::string frame_id;
+
+    /** ID of the object frame defined by this pose */
+    std::string object_frame_id;
+
+    /** Pose of the object frame in the reference frame */
+    base::TransformWithCovariance transform;
+
+public:
+    /** Default constructor */
+    PoseWithCovariance();
+
+    /** Initializes type from a TransformWithCovariance */
+    explicit PoseWithCovariance(const base::TransformWithCovariance& transform);
+
+    /** Initializes type from a RigidBodyState */
+    explicit PoseWithCovariance(const base::samples::RigidBodyState& rbs);
+
+    /** Performs a composition of this pose with a given pose.
+    * The result is another pose with result = this * pose.
+    * The frame ID's will be set accordingly (e.g. a_in_b = c_in_b * a_in_c).
+    * Note that this method doesn't check if the composition
+    * is frame wise valid.
+    */
+    PoseWithCovariance operator*(const PoseWithCovariance& pose) const;
+
+    /** Sets the transformation as TransformWithCovariance */
+    void setTransform(const base::TransformWithCovariance& transform);
+
+    /** Sets the transformation as Eigen::Affine3d */
+    void setTransform(const Eigen::Affine3d& transform);
+
+    /** Returns transformation as TransformWithCovariance */
+    const base::TransformWithCovariance& getTransformWithCovariance() const;
+
+    /** Returns transformation as Eigen::Affine3d */
+    Eigen::Affine3d getTransform() const;
+
+    /** Returns the 6x6 covariance matrix */
+    const base::TransformWithCovariance::Covariance& getCovariance() const;
+
+    /** Converts the type to a RigidBodyState.
+     * Note that the cross-covariances will be lost,
+     * since the RigidBodyState does not store them.
+     */
+    base::samples::RigidBodyState toRigidBodyState() const;
+};
+
+}}
+
+#endif
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -30,6 +30,7 @@ rock_library(
         samples/Sonar.cpp
         samples/SonarBeam.cpp
         samples/SonarScan.cpp
+        samples/PoseWithCovariance.cpp
     HEADERS
         Angle.hpp
         CircularBuffer.hpp
@@ -79,6 +80,7 @@ rock_library(
         samples/Sonar.hpp
         samples/SonarBeam.hpp
         samples/SonarScan.hpp
+        samples/PoseWithCovariance.hpp
         samples/Wrench.hpp
         samples/Wrenches.hpp
         templates/TimeStamped.hpp
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -32,6 +32,7 @@
 #include <base/samples/DepthMap.hpp>
 #include <base/TransformWithCovariance.hpp>
 #include <base/samples/Sonar.hpp>
+#include <base/samples/PoseWithCovariance.hpp>
 #include <base/Temperature.hpp>
 #include <base/Time.hpp>
 #include <base/TimeMark.hpp>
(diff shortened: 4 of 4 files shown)


base/samples/Transformation.hpp
```diff
@@ -0,0 +1,90 @@
...
+ * \f$ ^{world}_{sensor}T = ^{world}_{body}T \cdot ^{body}_{sensor}T \f$
+ */
+class Transformation
+{
+public:
+    /** Reference timestamp of the transformation sample */
+    base::Time time;
+
+    /** ID of the global reference frame */
+    std::string global_frame_id;
```
doudou: This naming scheme is, to me, a LOT more confusing than RigidBodyState's convention (and that says something, given how I fought to understand RBS's convention). Which means that there's really no confusion (or, more likely, that confusion is unavoidable). What there is, is lack of documentation in RBS.

A transformation is not between a "global" and "local" frame. A 6D transformation /transforms/ a valid pose expressed in one frame into a valid pose expressed in another frame (that's the RBS convention)

Or a transformation is the pose of a frame into another (which is I guess what you call 'local' and 'global', but nothing's global here ... I have already a hard time explaining the fact that transformations are purely relative things to not have the types adding more to the confusion).

If you really want to go for the latter, then I would personally prefer

```
std::string frame_id;
std::string reference_frame_id;
```

chhtz: The motivation for the naming was that "global" is "more global", but I agree that the naming is not perfect. I don't think ``reference_frame_id`` will be entirely clear either (I assume you mean the "more global" frame).
How about ``base_frame`` and ``local_frame`` or ``parent_frame`` and ``child_frame``?
O.t.o.h., if it is mostly a documentation thing, we could stick with the RBS nomenclature but make the documentation more clear.

We should actually also make it more clear (and more consistent) in what frame the covariance information is stored.

Besides the nomenclature, another intention of ``Transformation`` is to have a type which represents the full covariance (with cross-covariances) of 3D transformations while not having to store velocity information (which often is not needed/not available).

Regarding the docu of RBS: I always assumed that it expresses the Pose of the "source" inside the "target", i.e., when extracting an ``Eigen::Transform T`` of the RBS and I have a vector ``x_source`` in the source frame, I can calculate ``x_target=T*x_source``. (At least that is how ``operator Eigen::Transform<...>`` is implemented).
The documentation however says that the orientation is the orientation of the target frame inside the source frame, whereas the position is the position of the source frame inside the target frame (only the latter is consistent with the implementation).
For velocities it is inconsistent again: translational velocities are expressed in the target frame, whereas angular velocities are expressed in the "body fixed frame" (which I assume means the source frame).
But as there are no interactions between velocities and the Pose (inside the implementation of RBS), this is likely handled differently by different users of RBS ...

jmachowinski:
I agree, global and local is really confusing. parent / child is a bit better, but is more or less
equal to source / target. So why not stay with that naming scheme ? Or do I miss an imported
drawback of this scheme ?

saarnold:
I'm absolutely fine with changing the frame names to something else. But I really would like to avoid source and target, since they are ambiguous. One could read it as a transformation transforming a vector v from the source to the target frame: v_target = T * v_source (Which would be the right thing to do). But one could also read it as a translation vector pointing from the source frame to the target frame (which would be the inverse). And I know that this is confusing for a lot of people.

The goal should be to find a more distinct naming scheme. What about just `frame_id` and `child_frame_id`?

chhtz:
I think just `frame_id` should be avoided. And if we use it, I would rather use `frame_id` and `parent_frame_id`, since the child frame is usually the more distinctive one.

That said, I don't have too strong opinions on the naming, just let us not [bikeshed](http://de.urbandictionary.com/define.php?term=bikeshedding) this ...

saarnold:
Since the transformation would be expressed in the `frame_id` the other frame should be called `child_frame_id` IMHO.

> That said, I don't have too strong opinions on the naming, just let us not bikeshed this ...

:+1:

doudou:
> Since the transformation would be expressed in the frame_id

See, that's what I am talking about. A *transformation* is not expressed in *any frame*.

A transformation is actually an ill-defined term by itself. It is either the pose of a "frame" (i.e. the reference body of a frame) into another (which was the reason why I proposed the reference_frame/frame terminology - the pose of 'frame' expressed in 'reference_frame'), or a frame change operation that changes poses expressed in frame A into poses expressed in frame B (the source/target terminology)

base/samples/Transformation.hpp
doudou: > Besides the nomenclature, another intention of Transformation is to have a type which represents the full covariance (with cross-covariances) of 3D transformations while not having to store velocity information (which often is not needed/not available).

History: the reason why RBS exists as a pose+velocity is that *if* you have a pose+velocity estimator using a RBS, that estimator will not* be able to be connected to a consumer of a Transformation. Since having a pose+velocity estimate is not common but not rare either, we packed them in the same type.

RBS represents a frame change operation, which changes poses (or subparts of it) expressed in the source frame into the equivalent expressed in the target frame.

Which means that the pose part of it is the pose of the source in the target

The logical extension of this (since a RBS is the pose of A in B) is that the velocity part should be the velocity of A in B as well.

doudou:
> It is either the pose of a "frame" 

OK...

Why not calling this type Pose ? That would IMO remove all the ambiguity.

And the frame_ids could become object_id/frame_id (as "pose of object in frame")

chhtz:
No objections against calling it `Pose`. I usually would use a `Pose` as a transformation, but if `Pose` sounds less ambiguous, I'm fine with that.

saarnold:
I would be fine with Pose as well. But I really don't like object_id, we could call it object_frame_id.

jhidalgocarrio:
IMO. I agree "global" and "local" frame does not improve much (just a bit) the existing "source" and "target" which **confuse** to the user about the transformation to write.

Now. My proposal is to merge the PR #86. We would like to have EnviRe in Rock (another topic) but it solves the problem of producing samples (sensory data) coming from one sensor (produced in a frame "frame_id" of the sensor) at a particular time (base::Time).

Then, we can have a *Transformation* which has the SpatialTemporal information to set the transformation origin. It set "when and where" the transformation was produced. Something like:
```
SpatioTemporal<Pose> transformation;
```
where Pose has only:

```
std::string object_frame_id;
```
In this way, there is not room to confuse the user. The originating frame is the *frame_id* because it comes from the SpatioTemporal type. The child frame is the *object_frame_id*.
The question of naming this class  *Transformation* or *Pose* does not change much in my opinion.

base/samples/Transformation.hpp
```diff
@@ -0,0 +1,90 @@
...
+ * Example of a transformation chain:
+ * \f$ ^{world}_{sensor}T = ^{world}_{body}T \cdot ^{body}_{sensor}T \f$
+ */
+class Transformation
+{
+public:
+    /** Reference timestamp of the transformation sample */
+    base::Time time;
+
+    /** ID of the global reference frame */
```
Rauldg: I would change this comment (line 30) to one more illustrative like "The Id of the frame from which the transformation is defined (a.k.a. source, from, origin, reference)" and same applies for comment in line 60, where the "local_frame_id" is explained.

doudou:
> Now. My proposal is to merge the PR #86

My problem with this spatio-temporal adapter is that it embeds metadata (the frame IDs) in the data. Which makes changing the system after the fact a LOT harder: for instance when replaying data, dynamically rewiring components, or in multi-robot contexts where all these names must be *globally* unique.

I *know* that is is a sad fact of the Rock life right now because unfortunately (1) there is no support for other metadata channels and (2) RBS already has it (with the associated pain of having every single transform producer/consumer needing to know about frames even though most of them really do not care).

On top of the rest, std::string is not realtime-compatible.

A.k.a. PR #86 IMO expands the main problem of RBS, the frame names, eventually to most other types even though *right now* there is little need to. It already spread to Joints ... I would rather avoid a generalization.

base/samples/Transformation.hpp
doudou: > Regarding the docu of RBS

Honestly, if the docu of RBS looks weird, I would personally assume that the problem is that it probably does not match the actual use of RBS. All RBSes I've presonally seen do represent the pose of source in target. For velocities, my experience is that there are as many conventions as people actually generating RBSes.

doudou:
> chhtz: No objections against calling it Pose. I usually would use a Pose as a transformation, but if Pose sounds less ambiguous, I'm fine with that.
> saarnold: I would be fine with Pose as well. But I really don't like object_id, we could call it object_frame_id

So .. we were getting somewhere it seems.

Since Pose is taken, and since there is already base::PoseWithUncertainty, why not:

 - ` base::samples::PoseWithUncertainty`
 - `object_frame_id` / `frame_id`

?

saarnold:
Sorry for the delay, I had it always on my list just not with the highest priority I must admit.

I did apply the changes we discuses and adapted the documentation.

> Since Pose is taken, and since there is already base::PoseWithUncertainty

I couldn't find this type. I do like `PoseWithUncertainty` more but the type is now named `base::samples::PoseWithCovariance` since we already have a `TransformWithCovariance` and a `TwistWithCovariance`.

src/samples/PoseWithCovariance.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    /** Pose of the object frame in the reference frame */
+    base::TransformWithCovariance transform;
+
+public:
+    /** Default constructor */
+    PoseWithCovariance();
+
+    /** Initializes type from a TransformWithCovariance */
+    PoseWithCovariance(const base::TransformWithCovariance& transform);
```
doudou: These two conversion constructors should probably be `explicit`

saarnold: Yes I agree.

doudou:
Looks good to me. Anyone else ?

jhidalgocarrio:
> I couldn't find this type. I do like PoseWithUncertainty more but the type is now named base::samples::PoseWithCovariance since we already have a TransformWithCovariance and a TwistWithCovariance.

I found more meaningful PoseWithCovariance since Covariance is one way to represent Uncertainty. This is why I used TransformWithCovariance.

**PoseWithCovariance is not RBS.**
I suggest to remove the method PoseWithCovariance::toRigidBodyState() to copy to RBS. Two reasons: 1) Already interpreting frames using the RBS convention.It forces the user to use this interpretation 2) Lineal and Angular Velocities are not part of PoseWithCovariance. RBS will be incomplete anyway.

saarnold:
> Already interpreting frames using the RBS convention.It forces the user to use this interpretation

I think that is exactly the reason why it is good to have this helper. It might be necessary to convert between those two types at every part of the code where the RBS is currently used as PoseWithCovariance. So why not having this helper method implemented once with the correct assignment of frames?

jhidalgocarrio:
> So why not having this helper method implemented once with the correct assignment of frames?

The helper method makes sense between compatible times. It makes more sense between BodyState (already existing type) and RBS, since both classes have pose and velocity information.

The helper method makes to depend on (include) RBS header. We don't have a completely independent alternative to RBS.

Having this helper method is the same as calling the frames source and target frames. Indirectly, yes, but again!!

jhidalgocarrio:
For the rest I am good to merge this PR :+1: but removing the helper method toRigidBodyState()

doudou:
@jhidalgocarrio I don't agree. RBS is *designed* to be incomplete. A RBS with only pose and orientation information *is* a valid RBS. You guys are going to hit yourselves in the foot in a major way later down the line if you don't realize that whenever you will want to propagate a full pose+velocity state, having two ports is going to be a major pain.

That's why RBS has all these fields. Not having a RBS-like type is going to be problematic.

jhidalgocarrio:
> You guys are going to hit yourselves in the foot in a major way later down the line if you don't realize that whenever you will want to propagate a full pose+velocity state, having two ports is going to be a major pain.

I am not saying we don't need a pose+velocity state type to send with a single port. This is a different topic and it is not PoseWithCovariance

> So why not having this helper method implemented once with the correct assignment of frames?

I think is good to have PoseWithCovariance independent of RBS, but it is a minor issue.

Ok. Go for this. :+1:

src/samples/PoseWithCovariance.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * Represents a 3D pose with uncertainty of the frame 'object_frame_id' in the frame 'frame_id'.
+ * A frame ID shall be globally unique and correspond to a coordinate system.
+ * Following the [Rock's conventions](http://rock.opendfki.de/wiki/WikiStart/Standards)
+ * coordinate systems are right handed with X forward, Y left and Z up.
+ *
+ * The pose \f$ ^{reference}_{object}T \f$ in this structure transforms a vector \f$ v_{object} \f$
+ * from the object frame to the reference frame:
+ * \f$ v_{reference} = ^{reference}_{object}T \cdot v_{object} \f$
+ * Example of a transformation chain:
+ * \f$ ^{world}_{sensor}T = ^{world}_{body}T \cdot ^{body}_{sensor}T \f$
```
jhidalgocarrio: What about adding an operator * ? It implements this operation. It uses the already existing operator * from TransformWithCovariance, set the frame id correctly according to this comment line and take the newest timestamp to the resulting PoseWithCovariance.

saarnold: An method doing this would need to throw an exception if the transformations are incompatible or frames are undefined. I'm not sure if such an operator would be so nice to use.

jhidalgocarrio: > An method doing this would need to throw an exception if the transformations are incompatible

Yes. It can be easily done by comparing "body" of ^{world}_{body}T is equal to "body" of ^{body}_{sensor}T in this given example .

```
t1.object_frame_id.compare(t2.frame_id);
```
The rest would be to call the TransformWithCovariance operator*, set the timestamp and the frame names.

This method can be very helpful.

saarnold: My point was about the usability of such a trivial method with the necessity to throw, not it's implementation. 

Since you think it would be helpful I implemented it. I was about to implement a `inverse()` method as well, but since for TransformWithCovariance the result of  `tr * tr2.inverse()` is covariance-wise unequal to `tr.compositionInv(tr2)` I think it would be misleading or even dangerous. I'd rather implement `compositionInv` and `preCompositionInv` as well, but I don't see much benefit in forwarding all those methods.

doudou: >  I don't see much benefit in forwarding all those methods.

Me neither. At this stage, I would really much rather prefer keeping the interface simple, as it was originally.

As for the frame checks, it's IMO a bad idea to start injecting these kind of checks in a type like this, which is not meant for general computation (again, AFAIK). If one wants proper typecheck-level frame checking, then one should go and try KUL's semantic geometry (or, at least, attempt at doing a clean compile-time implementation of it that does not entail doing string comparisons all the time).

jhidalgocarrio: Thanks for including the operator *. IMO is required because rock is handling transformations (not only transporting them) and the method might be very useful.

> I don't see much benefit in forwarding all those methods.

I agree.  IMO it will be overdoing it.  

I don't have further comments. :+1:

saarnold:
Are there any other concerns or can we merge this?

src/samples/PoseWithCovariance.cpp
```diff
@@ -0,0 +1,85 @@
...
+    frame_id(rbs.targetFrame), object_frame_id(rbs.sourceFrame), transform(rbs.position, rbs.orientation)
+{
+    transform.cov << rbs.cov_position, Eigen::Matrix3d::Zero(),
+                        Eigen::Matrix3d::Zero(), rbs.cov_orientation;
+}
+
+PoseWithCovariance PoseWithCovariance::operator*(const PoseWithCovariance& pose) const
+{
+    // check frame ID's
+    if(this->object_frame_id.empty() || pose.frame_id.empty())
```
doudou: Please don't require the frames to be set. I very much plan to get rid of all of these frames.

Note that empty frames are equal so you just have to remove the first if.

doudou:
Sorry, I missed new operator *. One comment there.

I very much plan to get rid of the whole "having explicit frames" in the data types in my systems. You're obviously free to continue with that, but if at all possible I would prefer not requiring them to be set.

src/samples/PoseWithCovariance.cpp
saarnold: Fine with that, but I would rather remove all the checks and leave it to the user

doudou: > I would rather remove all the checks and leave it to the user

Totally :+1:

saarnold: Then again checking the frames was kind of the hole point of this operator I guess.
So yeah I'll remove the first check.

saarnold: Between my comments I re-read what the reason was for this operator. 
I didn't like the frame checks and the resulting necessity to throw back then and since I'm not the only one I removed the checks. Sorry @jhidalgocarrio

Rauldg:
To summarize, what needs to be done to merge?
According to the discussion I think that it only to allow the usage of the type without the requirement of providing/using frame ids.
Can't the type already be used without setting these?

doudou:
I have nothing on my side ... @saarnold @jhidalgocarrio ?

saarnold:
From my side this can be merged as well.

doudou:
@jhidalgocarrio ?

jhidalgocarrio:
I am good with it. No further comments as mentioned before [here](https://github.com/rock-core/base-types/pull/85#discussion_r95327972)
IMO we can merge.

PR #86, opened 2016-06-29
Spatio temporal type

saarnold:
This type can be used to augment arbitrary data types with spatio-temporal information.
The main use-case for this type is to augment informations related to a environment representation (e.g. maps) or a geographic information system. Where type transport normally happens sporadicly and data types might not allow introspection.

This type will be directly supported by the new Envire.

### DIFF
--- a/base/SpatioTemporal.hpp
+++ b/base/SpatioTemporal.hpp
@@ -1,0 +1,71 @@
+#ifndef BASE_SPATIO_TEMPORAL_HPP
+#define BASE_SPATIO_TEMPORAL_HPP
+
+#include <string>
+#include <base/Time.hpp>
+#include <base-logging/Singleton.hpp>
+#include <boost/uuid/uuid.hpp>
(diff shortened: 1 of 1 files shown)


doudou:
If this adapter is needed mainly by envire, I would suggest leaving it there. At least until we have a better idea of its general usefulness.

Adding metadata to transported types is not trivial, as it makes replay-ability a lot harder, and more generally goes against the ability for change.

saarnold:
It would be possible to keep it in Envire for now yes.
However I don't see much harm in having it in base types and it would probably be easier to add support to other libraries.

doudou:
> However I don't see much harm in having it in base types and it would probably be easier to add support to other libraries.

IMO base-types should not be a collection of "types that might be useful". I would avoid having such a type here until it shows more usefulness than envire (and I do see harm in such a type in general, so that's also a reason why I definitely :-1: its inclusion in base types)

saarnold:
The type will be exported by envire.

PR #87, opened 2016-07-01
Depthmap: Corrects some calculations.

Rauldg:
Corrects for the polar case the calculation of the resolution and the horizontal angles from the min and max angle.

Thanks to @chhtz


saarnold:
@Rauldg thanks for reporting this. There is a problem when an angular range defined by two angles and is bigger then PI.
However your fix introduces other issues. It isn't possible to define an angular range from e.g. -90 to 90 degree while keeping the requirement that the data is ordered from the left to the right.
I opened a new PR which properly fixes this issue and also adds unit tests for this case:
rock-core/base-types#88

jakobs:
Can this PR be closed then?

PR #88, opened 2016-07-04
DepthMap: fixed an issue when computing the step resolution in the polar case

saarnold:
Thanks @Rauldg for reporting this issue.

It also add missing unit test for this case.

### DIFF
--- a/src/samples/DepthMap.hpp
+++ b/src/samples/DepthMap.hpp
@@ -71,13 +71,13 @@ public:
     std::vector<base::Time> timestamps;
     
     /** Defines the vertical projection type of the depth map.
-     * If polar, the vertical intervals are angles in the range of (-PI, PI].
+     * If polar, the vertical intervals are angular rotations around the Y-unit axis.
      * If planar, the vertical intervals are positions on an image plane coordinate frame.
      */
     PROJECTION_TYPE vertical_projection;
     
     /** Defines the horizontal projection type of the depth map.
-     * If polar, the horizontal intervals are angles in the range of (-PI, PI].
+     * If polar, the horizontal intervals are angular rotations around the Z-unit axis.
      * If planar, the horizontal intervals are positions on an image plane coordinate frame.
      */
     PROJECTION_TYPE horizontal_projection;
@@ -87,7 +87,9 @@ public:
      * In planar projection mode the vertical intervals are y coordinates on an 
      * depth-image plane, with zero in the middle of the plane.
      * In polar projection mode the vertical intervals are angular rotations
-     * around the Y-unit axis.
+     * around the Y-unit axis. Vertical angles must always be ordered from a smaller
+     * to a higher value, since the rows of the data matrices are interpreted from the upper
+     * to the lower row.
      * The field has either two or |vertical_size| entries. In the case of two 
      * entries the intervals are interpreted as upper and under boundaries. The 
      * transformation for each measurement will be interpolated.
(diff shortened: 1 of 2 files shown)


chhtz:
This does also fix the issue which was fixed in PR #87.
I don't know, if we want to allow `DepthMap`s which turn more than a full revolution (which could be helpful, when the robot mounting the LIDAR is turning while the LIDAR records, but one wants to be sure to actually have a full cycle).

Rauldg:
IMO, the sample should finish as soon as the sensor has covered the given range, according to its internal initial state, additional image processing should be performed latter. Otherwise, to take samples the sensor task would need the position of the robot and that makes it more complicated for simple use cases.

saarnold:
By using two angles to define a range having a constant angular resolution, this is not possible. We would need to add more meta information, which would overload this type even more and I'm against that. (One still could encode it by just using the non-discrete form)

In your use-case you would need to know how you have moved in the component producing the DepthMap, in this case you could simply transform the measurements before sending the full cycle.

chhtz:
It would not be possible by providing two `base::Angle`s, but by giving two absolute angles (which include the number of revolutions), e.g., give start angle 0 and end angle 10pi, to encode 5 full revolutions. Alternatively, one could provide the starting angle and the angle step, instead of the start and end angle. 
But admittedly, that can just as easy be encoded by 5 successive DepthMaps. So I'm ok with the PR as it is.

jakobs:
There is the AngleSegment class in the Ange.hpp file. Does this have any impact on your discussion?

chhtz:
I guess `AngleSegment` would make sense if `DepthMap` was always used with start and end-angle. Currently, there is an option to either specify every angle or to specify start and end-angle.

jmachowinski:
Angle segment for start/end is undefined. Or you would need to assume a direction.

jmachowinski:
I just remembered, the main problem was that you could not define a full turn with it.

Rauldg:
Any reason not to merge?

jmachowinski:
looks good to me

chhtz:
Overall I'd still prefer storing non-normalized angles here. This would allow to safely define an interval [0, 2π] (or [-π, +π]) without having to worry that it ends up to be [0, ε] (where ε is from a rounding error when normalizing 2π). 
Essentially, in `computeResolution` just replace the `base::Angle` by `double` and optionally make an assertion that `.front()` is smaller than `.back()` (or the other way around).

saarnold:
> This would allow to safely define an interval [0, 2π] (or [-π, +π]) without having to worry that it ends up to be [0, ε] (where ε is from a rounding error when normalizing 2π). 

Yes this could be an issue.

> Essentially, in computeResolution just replace the base::Angle by double and optionally make an assertion that .front() is smaller than .back() (or the other way around).

Currently both [0.5π, -0.5π] and [-0.5π, 0.5π] are valid ranges, one to the front and one to the back. With the constraint that the data is ordered row major from the left to the right.
Changing this is possible, but would mean that the range to the back needs to be defined [-0.5π, -1.5π].
Then [-0.5π, 0.5π] for the horizontal case would mean that the data is ordered from the right to the left instead.
In the hope that no one was using it that way we can change this.

chhtz:
I agree, my only concern with my suggestion would be that we break existing code as well.

From the point of usability I think  `[-0.5π, -1.5π]` (if it goes clockwise?) is much more intuitive than knowing that `[-0.5π, 0.5π]` actually goes `-0.5π --> -π == +π --> 0.5π`. Furthermore, with proper assertions I think almost all wrong parameters should get caught.

saarnold:
Ok essentially we are back to the origin PR now :)
I updated the documentation and unit tests.

> Furthermore, with proper assertions I think almost all wrong parameters should get caught.

We could throw an invalid argument exception when angles are ordered differently to the expected direction. But since it wouldn't break the general function of the type I would still leave that to the user and don't be too restrictive here.
According to the documentation there should be only one correct way of course.

saarnold:
Ping! From my side this can be merged.

PR #90, opened 2016-07-26
add missing Requires: on base-logging for the header-only parts

doudou:


jakobs:
I am wondering if it wouldn't be better to remove the base logging dependency instead. The only occurrence is in a test that should be moved anyway, and the use of the singleton class for a LUT in the depthmap, which makes me wonder if this is not somehow going beyond what you would expect in a base type. @saarnold @chhtz

chhtz:
Regarding the complexity of `DepthMap`: I asked @saarnold some time ago why the type includes all the conversion routines and I think he said that decision was made back than essentially because there would be no better place for base-type conversions (my memory might be wrong or I may have misunderstood him).

That said, I'm generally all for reducing dependencies and I think it makes sense to factor out complex conversion routines (especially, if we have a place for that, which is more or less obvious to get found).

jakobs:
I just realised that removing the dependency on base-logging from base-types would mean touching nearly all packages. Sorry for the noise, then. But will move the tests to the logging library.

PR #94, opened 2016-08-09
Splitted base/types to seperate cpp and hpp files

hemker:
We splitted the base-types header-only lib into a separate .cpp/.hpp implementation due to very long orogen parsing times. The old implementation needed about 8 minutes (on a SSD) to finish the orogen phase in base/orogen/types. With the new concept, an enhancement of nearly 6 minutes (parsing times around 2 minutes) can be achieved. Furthermore, we removed some deprecated stuff.

Split Pose.hpp into implementation and header

Float.hpp: Use std function, not boost ones

Eigen: remove isnotnan and isfinite

This is a duplicate of the buildin
allFinite()

move guaranteeSPD into TwistWithCovariance.

This one includes SVD, which is a big include in the base eigen.

moved Wrench.hpp to new src

moved TwistWithCovariance.hpp to new lib/src

moved Waypoint to new src folder

re-patched new isnan interface

splitted TransformWithCovariance

splitted Timueout

moved headers to cpp files

splitted TimeMark

removed inlining

splitted Time

splitted Temperature

moved Point

splitted Pressure

fixed compile errors

Added correct PC file for base-types, made base-lib a dummy

moved CircularBuffer

moved Deprecated

moved JointLimitRange

moved JointLimits

moved JointState

moved JointTransform

moved JointsTrajectory

moved NamedVector

moved Odometry

added odometry

ported rest of types

re-removed vector included

changed dep to right lib name

compile fixes


jakobs:
Would be great if a few people could test this, since this is a fairly big change.

jmachowinski:
Tested, works.
Some packages reported now missing headers, but this is actually a positive side effect.

jakobs:
One Problem I have with this PR, is that it is a single commit. It would be much better to have multiple commits, since there seem to be different changes involved. Also, (I havent' checked) it would be important imho to have the original authorships preserved.

chhtz:
Just glancing over it, I'd also prefer having it split into smaller commits.

At least one flaw: You copied the old/broken `guaranteeSPD()` which I recently fixed (https://github.com/rock-core/base-types/pull/93)

About `std::isfinite`, etc: Is rock-master officially >=C++11-only? (I would not object, just wanted to ask if that is official policy).

In the cpp-files, you could embrace everything into `namespace base { ... }` instead of prefixing everything with `base::`

Also, I'd prefer keeping the very simple getter/setter functions inlined in the header (I did not benchmark that, but I'd guess some will have measurable impact on runtime).

hemker:
I unsquashed the commits and  moved guaranteeSPD to TwistWithCovariance, the only file where it's used. It allowed me to remove a header from Eigen.hpp as well.

doudou:
> Tested, works.

Did someone actually did a *clean build* from a *clean machine (a.k.a. VM / container)* ? And, no, we have not switched master to C++11 yet. So if we want to keep the C++11 constructs, we must resolve the C++11 PR *first*.

And test that there are no undefined symbols in the generated library ? (with `ldd -r`) ?

doudou:
Moreover, this makes SISL a required dependency, it was only optional before.

doudou:
> The old implementation needed about 8 minutes (on a SSD) to finish the orogen phase in base/orogen/types

Do you have a Pentium II hooked on a SSD ? I was so surprised that I ran the test here, and it took 3 min (using castxml)

(Note: not that I object to the change, there's so much code in these headers nowadays that it probably makes sense for maintenance reasons, but parsing times ? Come on !)

hemker:
I tested it on my notebook, equipped with a SSD and a two year old i5...
Resulting castxml xml files were even larger than 8 MiB.

jakobs:
Performance test results on my insanely high-spec machine for building the orogen/types (orogen+compile) with castxml.
old master: 6m32.096s
split headers: 3m27.039s

@doudou: how did you get 3 mins on the old master?
@hemker: how did you get 2 mins with the new branch?

One more thing to note:
old: built base/orogen/types (124 warnings)
new: built base/orogen/types (149 warnings)

doudou:
> Performance test results on my insanely high-spec machine for building the orogen/types (orogen+compile) with castxml.
> old master: 6m32.096s
> split headers: 3m27.039s
> @doudou: how did you get 3 mins on the old master?

I only measured the orogen phase, since that was the claimed reason for the change.

doudou:
> on a fresh build.

GIven how critical that package is, it should really be a fresh build /on a fresh machine/

doudou:
And still two outstanding problems
 - C++11 dependency
 - SISL is not optional anymore

jakobs:
But compilation was quick compared to orogen phase (around 30s). But did you get 3min on the old master branch? Because that would really be interesting what is different in your setup.

chhtz:
How about as a first step just remove the deprecated stuff and only put some of the heavy functions into cpp-files?

doudou:
> But compilation was quick compared to orogen phase (around 30s). But did you get 3min on the old master branch? 

Yes. I'll try to get some time to test with a vanilla setup (i.e. without the piles of branches and unmerged PRs I am on)

jakobs:
When compiling from a fresh checkout, I get:

In file included from /home/jakob/dev/rock/base/types/src/samples/CompressedFrame.cpp:1:0:
/home/jakob/dev/rock/base/types/src/samples/CompressedFrame.hpp:7:2: warning: scoped enums only available with -std=c++11 or -std=gnu++11
  enum class frame_compressed_mode_t {
  ^
/home/jakob/dev/rock/base/types/src/samples/CompressedFrame.cpp: In static member function ‘static base::samples::frame::frame_compressed_mode_t base
::samples::frame::CompressedFrame::toFrameMode(const string&)’:
/home/jakob/dev/rock/base/types/src/samples/CompressedFrame.cpp:6:16: error: ‘frame_compressed_mode_t’ is not a class or namespace
         return frame_compressed_mode_t::MODE_COMPRESSED_UNDEFINED;
                ^
/home/jakob/dev/rock/base/types/src/samples/CompressedFrame.cpp:8:16: error: ‘frame_compressed_mode_t’ is not a class or namespace
         return frame_compressed_mode_t::MODE_PJPG;
                ^

hemker:
@jakobs I only measured the orogen phase time. Considering compile phase, durations about 3 minutes are reached as well. The error with the enum class is a redefinition error of MODE_PJPG. It only occurs when moving the use of the enum into a separate cpp file. The enum value (better: the symbol) is first defined in Frame.hpp. Using c++11 it allows to change the enum to an enum class - it's a new feature that puts the enum values into the right, non-redieining scope.

doudou:
> The error with the enum class is a redefinition error of MODE_PJPG

Avoid changing that stuff within this PR please. Stick to moving code around, while keeping the actual organization the same.

You can always propose this kind of changes later on in separate PRs.

hemker:
but the code won't compile then

doudou:
> but the code won't compile then

The existing code *did* compile. Unless you changed it in major ways (which you should defnitely not do in this PR), there is way to do a hpp/cpp split and keep C++98 compatibility.

jakobs:
The PR as it it does not compile on my machine on a new bootstrap (due to the c++11 features).

hemker:
So remove all c++11 related features or rather make it c++98 compliant?

doudou:
> So remove all c++11 related features or rather make it c++98 compliant?

Yes. C++11 is still not enabled globally on rock (it's in the works, but not there), and that's really not the point of this pull request. Stick to splitting headers and cpp files here, and come back with the introduction of the C++11 features in a separate PR.

hemker:
I completely removed the deprecated CompressedFrame class. It wasn't able to compile before because no other module ever included the CompressedFrame.hpp and caused a redefinition error. This class was the only one making use of c++11 features and was removed now.

hemker:
I finally moved all namespaces into namespace { } style.

jmachowinski:
Looks good and compiles on my machine.

jmachowinski:
On a clean master bootstrap (non c++11) there were 2 includes missing.
Also base/orogen/types needs https://github.com/rock-core/base-orogen-types/pull/23

jmachowinski:
There are also some warning about base::Temperature base::operator- but we can leave them for another pr. From my side, it would say merge.

jakobs:
I still have some issues. Your search/replace removing base:: was a bit too thorough. It changes all the type string in exceptions, and there was one occurence where ios_base was replaced with ios (I'll mark it).

jakobs:
And I was wondering why the ios_base replacement wasn't picked up. That is of course because the tests are not building by default anymore. You have to do 
autoproj test --enable
amake
Even the cmake configure is broken with this PR. Tests on master are working fine.

hemker:
@jakobs I made the tests compile again. But there's one inconvenience: Since we moved the method guaranteeSPD from base/Eigen.hpp to TwistWithCovariance.cpp - because it's the only class where it is used - the test suite test_Eigen.cpp cannot compile anymore. What to do now? Remove test_Eigen.cpp? Or make the guaranteeSPD an interface function of TwistWithCovariance. But that would cause to put the templated method into the header again and make it public.

jmachowinski:
It's a bit hacky, but include the cpp file in the test. That should work fine.

hemker:
Done. Works fine

chhtz:
This of course makes the declaration of `guaranteeSPD` in base/Eigen.hpp essentially useless. 
I'm not totally convinced that this really is a useful function -- considering that "everything worked fine"™ for very long with a broken implementation of that function. 
@jhidalgocarrio (or anyone else): Did using this function make a difference somewhere?

jhidalgocarrio:
Please, do not move the guaranteeSPD method to TwistWithCovariance.cpp
As @hemker suggested, make the guaranteeSPD an interface function of TwistWithCovariance and put the guaranteeSPD templated method into the header again and make it public.
The idea is to use the guaranteeSPD method for any arbitrary Eigen matrix.

chhtz:
We could leave the function in Eigen.hpp but hide the actual implementation in a function which accepts only `const Matrix<S,Dynamic,Dynamic>&`, implement that function in a cpp file and instantiate that only for `double`, `float`, `complex<double>` and `complex<float>`.

jmachowinski:
We could also leave it a template and move it into some other header. It was moved out of the Eigen.hpp header, as it adds a couple of includes.

jhidalgocarrio:
Both solution could work, but please do not move it to TwistWithcCovariance.
I guess @jmachowinski prefers guaranteeSPD out of Eigen.hpp for compilation time, but moving it to another header does not really solve the compilation time problem. It moves it to another header which needs to be included when using the method.
what @chhtz proposes is more reasonable to me +1. Meaning that this PR is all about "very long parsing and compilation time".

chhtz:
Moving it to a separate header (which is not included by Eigen.hpp) would make a difference, because a lot of packages are (indirectly) including Eigen.hpp (but don't need `guaranteeSPD`). It would require users of that function to add an `#include`, of course.

jhidalgocarrio:
Alright. I am also fine with it and it seems the easiest  :+1:

hemker:
I moved guaranteeSPD to it's own header, removed Eigen includes from TwistWithCovariance and made its .cpp  dependend on new Matrix.hpp

chhtz:
There was still the concern https://github.com/rock-core/base-types/pull/94#issuecomment-238507640 to keep the very trivial functions inside the header, i.e., at least the one-liner functions which are unlikely to change (and don't depend on including heavy external headers). This was also addressed on the mailing list.

Other than that, I don't have objections, though I did not actually test the most recent version.

Regarding orogen/compile time: This change will actually save a lot of additional time since at the moment (almost) every package needs to be rebuild even for small changes in function bodies.

jmachowinski:
> There was still the concern #94 (comment) to keep the very trivial functions inside the header, i.e., at least the one-liner functions which are unlikely to change (and don't depend on including heavy external headers). This was also addressed on the mailing list.

This is not critical and can be changed over time if needed. I would merge now, so that we don't collide any more with other merges going into base types.

src/Temperature.hpp
```diff
@@ -0,0 +1,113 @@
...
+    
+    bool operator==(const Temperature &other ) const;
+    
+    bool operator<(const Temperature &other ) const;
+    
+    bool operator>(const Temperature &other ) const;
+
+};
+
+static Temperature operator+( Temperature a, Temperature b );
```
jakobs: I believe this is not correct (and the compiler warns about it!). It was static inline before, which is fine since the header was always in the respective compilation unit. When declaring it, static means the function is local to this compilation unit, and will not get exported.

chhtz:
@jmachowinski All right, I'm fine with that (but I guess, I don't have the final call on this).

src/Temperature.hpp
chhtz: I think this should be declared as friend method inside the class. Alternatively, (also inside the class) as
```
Temperatur operator+(const Temperature & other) const;
```

jakobs: just removing the static should be ok. It was working with the original code.

doudou:
> All right, I'm fine with that (but I guess, I don't have the final call on this).

This is base-types. No one does have "the final call on this"

doudou:
And :+1: on Janosch's point. It's good enough for now, let's dwelve into C++11 / optimizations later.

jakobs:
Ok, I don't have any more objections. From my point this is good to merge.

src/Temperature.hpp
```diff
@@ -97,13 +97,13 @@ class Temperature
...
-static Temperature operator-( Temperature a, Temperature b );
+Temperature operator-( Temperature a, Temperature b );
 
-static Temperature operator*( Temperature a, double b );
+Temperature operator*( Temperature a, double b );
 
-static Temperature operator*( double a, Temperature b );
+Temperature operator*( double a, Temperature b );
 
 static std::ostream& operator << (std::ostream& os, Temperature temperature);
```
chhtz: I guess this `static` must be removed as well (this gives a `[-Wunused-function]` warning, too).

doudou: Especially because a static of a declaration looks absolutely wrong.... You can't be static without a body defined somewhere.

jmachowinski:
Compiles fine in fresh bootstrap and my own project

PR #96, opened 2016-08-13
Move LinearAngular6DCommand from auv_control

rafaelsaback:
@saarnold @doudou 

I've moved `LinearAngular6DCommand` from auv_control to base/types, since previously many other components had to depend on auv_control only to have access to this specific type.

I put the type into `base/commands` namespace and created a typedef inside `base` to guarantee backwards compatibility.

### DIFF
--- a/base/commands/LinearAngular6DCommand.hpp
+++ b/base/commands/LinearAngular6DCommand.hpp
@@ -1,0 +1,49 @@
+#ifndef BASE_COMMANDS_LINEAR_ANGULAR_6D_COMMAND_HPP_
+#define BASE_COMMANDS_LINEAR_ANGULAR_6D_COMMAND_HPP_
+
+#include <base/Time.hpp>
+#include <base/Eigen.hpp>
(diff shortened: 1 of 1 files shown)


saarnold:
:+1:

base/commands/LinearAngular6DCommand.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+            double x() const { return linear(0); }
+            double y() const { return linear(1); }
+            double z() const { return linear(2); }
+            double roll() const { return angular(0); }
+            double pitch() const { return angular(1); }
+            double yaw() const { return angular(2); }
+        };
+    }
+    typedef base::commands::LinearAngular6DCommand LinearAngular6DCommand;
```
doudou: Should add a comment above that states this is only for backward compatibility purposes.

rafaelsaback:
@saarnold @doudou I've added the comment, feel free to merge this PR

PR #97, opened 2016-08-19
remove the Eigen bindings

doudou:
They are moving to a separate package/gem

See https://github.com/rock-core/package_set/pull/81


doudou:
The two PRs need to be moved together. The package set PR got attention only from Janosch ... maybe this one will interest more ruby-oriented people ?

Issue #99, opened 2016-10-13
Angular velocity frame representation


joaobrittoneto:
The RigidBodyState's documentation says the RBS represents the state of the sourceFrame expressed in the targetFrame.
But the angular velocity is expressed in body fixed frame (I'd assume it's the sourceFrame).
Why it's not represented in targetFrame, like the velocity?

doudou:
@jhidalgocarrio : any idea ? How are the velocities represented currently w.r.t the source and target frames ?

jhidalgocarrio:
> But the angular velocity is expressed in body fixed frame (I'd assume it's the sourceFrame).

This is a wrong assumption (but it is not your fault). Rock naming convention is "IMO" a mess and I have been saying that since the last 3 years.
In Rock you have to think that the sourceFrame is not the fixed frame, it is the moving frame. "source" as the source of a sensor frame, which is normally the moving frame. (never interpret "source" as the source of the RBS transformation, as 90% of people do).

In RBS documentation is written (line26): 
"For instance, if sourceFrame is "body" and targetFrame is "world", then
the RigidBodyState object is the state of body in the world frame"

Then, I assume angular velocity is expressed in sourceFrame which is the moving frame.

> Why it's not represented in targetFrame, like the velocity?

It is actually expressed in "target frame". Rock calls it sourceFrame.

doudou:
>  Rock naming convention is "IMO" a mess and I have been saying that since the last 3 years.

I wholeheartedly agree on that part ... I'm trying to get people to at least *document* the thing properly so that one can make sense of it.

The source/target naming convention being based on the 'frame change' convention (i.e. X/source*RBS(source->target)=X/target), it means that RBS(source->target) is the pose of the source expressed in the target frame, right ? (given that Identity/source*RBS(source->target) = Pose_source/target.

If that's right, I get your moving frame / reference frame argument. Basically, if a RBS represents the state of body X in reference Y, sourceFrame = X, targetFrame = Y.

joaobrittoneto:
Thanks for the explanation.

>(never interpret "source" as the source of the RBS transformation, as 90% of people do)

I have to admit it was my case for a while.

So, If i got this right regarding the velocities:
The angular velocity is expressed in sourceFrame ( the angular velocity of the moving frame expressed on it own frame).
And the linear velocity is expressed in targetFrame (the linear velocity of the moving frame expressed on the fixed frame). Is that right? 

[...]

PR #100, opened 2016-10-20
Make explicit that body fixed frame is the sourceFrame

joaobrittoneto:
As the frames are formally described as sourceFrame and targetFrame,
make it clear that the angular velocity is represented in sourceFrame.
Let an indication that it is the "body" frame, as in the RigidBodyState documentation

### DIFF
--- a/src/samples/RigidBodyState.hpp
+++ b/src/samples/RigidBodyState.hpp
@@ -61,7 +61,7 @@ namespace base { namespace samples {
 	 */
         base::Matrix3d cov_velocity;
 
-        /** Angular Velocity as an axis-angle representation in body fixed frame
+        /** Angular Velocity as an axis-angle representation in sourceFrame (body fixed frame)
          *
          * The direction of the vector is the axis, its length the speed */
         base::Vector3d angular_velocity;
--- a/src/samples/RigidBodyState.hpp
+++ b/src/samples/RigidBodyState.hpp
@@ -55,13 +55,15 @@ namespace base { namespace samples {
 	 */
         base::Matrix3d cov_orientation;
 
-        /** Velocity in m/s of sourceFrame expressed in targetFrame */
+        /** Velocity in m/s of sourceFrame relative to targetFrame,
+         * expressed in targetFrame */
         base::Vector3d velocity;
 	/** Covariance of the velocity 
 	 */
         base::Matrix3d cov_velocity;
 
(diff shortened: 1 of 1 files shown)


doudou:
This goes against my understanding of https://github.com/rock-core/base-types/issues/99

If RBS is the state of 'source' in 'target', the velocities should be the velocities of 'source' in 'target'. Or am I still missing something ?

saarnold:
I agree that both velocities should be represented in the same frame. In most parts of the code I know it is sadly not the case, probably because of the ambiguous documentation.
Changing body fixed frame to world fixed frame would be one solution. 
I would also be more than happy to have both velocities represented in the source frame, but the effort of changing this might be even higher I guess.

saarnold:
At least the code I know, where both pose and velocity is set, is consistent with the current documentation. But it wouldn't be to hard to change those parts.

joaobrittoneto:
I have a doubt about the velocities in case the sourceFrame==targetFrame.
The velocities are not the velocities of the sourceFrame related to the targetFrame, expressed in the respectives frames? In this case the velocity of sourceFrame==targetFrame should not be zero??

doudou:
> In this case the velocity of sourceFrame==targetFrame should not be zero??

For me, the only thing that would make sens with the current definition is to express the velocity of sourceFrame in the targetFrame considered fixed at time (t). Otherwise, there is no way to express a velocity of a vehicle in the body frame.

joaobrittoneto:
>to express the velocity of sourceFrame in the targetFrame considered fixed at time (t)

It's not a trivial assumption. One can think that the frames, if having the same name, should share the same properties, including velocities. 

Why not represent the velocities direct in sourceFrame (the body frame)? Like the velocity of the sourceFrame related to the targetFrame expressed in sourceFrame?
Like that the information of orientation between sourceFrame and targetFrame doesn't really matter  at that level. 
For example, the velocities of sourceFrame related to two frames, targetFrame1 and targetFrame2, both expressed in sourceFrame, are the same if the transformation between targetFrame1 and targetFrame2 is constant.

A body frame velocity is usually what a sensor measures, independent of the targetFrame (inertial frame) used. If anyone wants the representation of the velocity in a specific targetFrame, then the orientation is required. 
And I have the impression that most of a system requires the velocities expressed in body frame. Having the velocity expressed in targetFrame requires the orientation being available or making the hack of sourceFrame==targetFrame.

doudou:
Thanks

PR #101, opened 2016-12-09
src: DistanceImage fix getPointCloud method

jhidalgocarrio:
the conversion from DistanceImage to PointCloud was creating PointClouds of uninitialized points.


doudou:
Would be great if you could write a unit test ...

jhidalgocarrio:
Unit test added.

PR #102, opened 2016-12-23
ruby: create a RBS.invalid value only once and dup it

doudou:
Just to avoid generating huge amounts of objects just to create
an invalid RBS


doudou:
Ping

PR #103, opened 2017-01-03
Fix of Eigen alignment exceptions with Depth Map

mmunoz-gmv:
Fix for several instances of an Eigen alignment exception:

test-range: /usr/include/eigen3/Eigen/src/Core/DenseStorage.h:86: Eigen::internal::plain_array<T, Size, MatrixOrArrayOptions, 16>::plain_array() [with T = double; int Size = 4; int MatrixOrArrayOptions = 0]: Assertion `(reinterpret_cast<size_t>(eigen_unaligned_array_assert_workaround_gcc47(array)) & 0xf) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"' failed.

In DepthMap.hpp: using Eigen::aligned_allocator in the declarations of std::vector containing Eigen vectors or matrices.

In DepthMapVisualization.hpp: adding the Eigen macro for the declaration of classes with Eigen vector/matrix members.

### DIFF
--- a/src/samples/DepthMap.hpp
+++ b/src/samples/DepthMap.hpp
@@ -239,8 +239,8 @@ public:
 	point_cloud.reserve(distances.size());
 	
 	// precompute local transformations
-	std::vector< Eigen::Transform<typename T::Scalar,3,Eigen::Affine> > rows2column;
-	std::vector< Eigen::Transform<typename T::Scalar,3,Eigen::Affine> > columns2pointcloud;
+	std::vector< Eigen::Transform<typename T::Scalar,3,Eigen::Affine>, Eigen::aligned_allocator< Eigen::Transform<typename T::Scalar,3,Eigen::Affine> > > rows2column;
+	std::vector< Eigen::Transform<typename T::Scalar,3,Eigen::Affine>, Eigen::aligned_allocator< Eigen::Transform<typename T::Scalar,3,Eigen::Affine> > > columns2pointcloud;
 	computeLocalTransformations(rows2column, columns2pointcloud, use_lut);
 	
 	// convert rows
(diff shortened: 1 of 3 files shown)


saarnold:
Hi @mmunoz-gmv thanks for the fix.
The vectors in line 312 and 342 should be changed as well.

mmunoz-gmv:
I've changed those vectors, and some others in DepthMap, LaserScan and their associated viz clases and test.

saarnold:
As far as I know it is not necessary for Eigen types that are not a multiple of 16 bytes, since they do not fit into common SIMD register sizes. Therefor it shouldn't be necessary for a vector of Eigen::Vector3d (192 bit) or Eigen::Vector3f (96 bit).
This is also true for the templated vectors in DepthMap, they need to be three dimensional.

mmunoz-gmv:
Another try...

saarnold:
Looks fine to me :+1: 
But I'm sure before this is merged someone will tell you to [squash](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits) your commits. It is good practice to have as much commits as necessary to describe what you changed.

jhidalgocarrio:
+1 for me too.

mmunoz-gmv:
OK, reduced to 2 commits.

src/samples/DepthMap.hpp
```diff
@@ -239,8 +239,8 @@ struct DepthMap
 	point_cloud.reserve(distances.size());
 	
 	// precompute local transformations
-	std::vector< Eigen::Transform<typename T::Scalar,3,Eigen::Affine> > rows2column;
-	std::vector< Eigen::Transform<typename T::Scalar,3,Eigen::Affine> > columns2pointcloud;
+	std::vector< Eigen::Transform<typename T::Scalar,3,Eigen::Affine>, Eigen::aligned_allocator< Eigen::Transform<typename T::Scalar,3,Eigen::Affine> > > rows2column;
```
doudou: Eigen types with alignment constraints should never be used on these types, as it breaks on any form of reflection
(e.g. typelib and RTT).

This is the reason for the base::* typedefs. Create base:: typedefs for the relevant types (transform and friends).

doudou: Disregard my comment .... just realized that it was within a method (and not as a field).

jhidalgocarrio: Is there something missing that prevents to merge this PR?

doudou: Not from my POV. As I said, my comments were completely irrelevant. Sorry for the noise !

jhidalgocarrio:
@doudou or @jmachowinski  when possible.  Could you merge it?

PR #104, opened 2017-01-12
viz: Waypoint visualization has not rock convention orientation of heading

jhidalgocarrio:
For some reason the waypoint visualization is artificially rotated -90deg to point X to the right. In rock convention X is forward.

### DIFF
--- a/viz/WaypointVisualization.cpp
+++ b/viz/WaypointVisualization.cpp
@@ -104,9 +104,7 @@ void WaypointVisualization::addWaypoints(osg::Group* group) {
         osg::Vec3 position = eigenVectorToOsgVec3(it->position);
         position[2] += 0.01; // Moves the waypoints a little bit above the z=0 plane.
         waypoint_transform->setPosition(position);
-        // osg::Quat(0,0,1,heading) != osg::Quat(heading, Vec(0,0,1)).. why?
-        // -M_PI/2.0 because rock defines x to be the front axis.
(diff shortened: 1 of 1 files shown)


doudou:
Hi @jhidalgocarrio. the branch also contains other unrelated things.

Generally speaking I would agree, but then how do we deal with backward compatibility ? Is that a problem that everyone has noticed ? I'm puzzled that you're the first to notice ...

jhidalgocarrio:
> Hi @jhidalgocarrio. the branch also contains other unrelated things.

My bad. Somehow I messed up with a merge from another open PR recently closed.  It should be fixed by now.

> Is that a problem that everyone has noticed ?

No problem so far. The code comes from iMoby times.

jhidalgocarrio:
> How do we deal with backward compatibility ?

My guess. It would not have compatibility with some old iMoby code, but I doubt this code is currently in used (I don't think it even compiles with the current base/types and base/logging).

IMO let be practical and implement the rock coordinate convention in base/types.

doudou:
Something I don't get. I had a look at the code, and the arrow is actually created pointing towards Y and then rotated. The comment (that you removed) actually points out that the rotation is *because* of Rock's convention.

If I visualize a WP with a heading of zero I get this

wpnav

A.k.a. what I would expect. Would you be mixing NWU / ENU conventions for the world frame ? Or some problems related to the world frame of your pose estimation ?

doudou:
Ping

doudou:
Given the state of the discussion, I'm closing this. Reopen if you have more comments.

PR #105, opened 2017-02-05
two new comparison operators to base/Angle

romulogcerqueira:
add less than or equal (<=) and greater than or equal (>=)

### DIFF
--- a/src/Angle.hpp
+++ b/src/Angle.hpp
@@ -169,6 +169,16 @@ public:
         return this->rad > other.rad;
     }
 
+    inline bool operator<=(const Angle &other ) const
+    {
+        return this->rad <= other.rad;
+    }
+
+    inline bool operator>=(const Angle &other ) const
+    {
+        return this->rad >= other.rad;
+    }
+
     inline Angle &operator+=(const Angle &other )
     {
         this->rad += other.rad;


romulogcerqueira:
ping

jmachowinski:
This looks a bit undefined.
Can you describe, how this operator should work ? 
Keep in mind that there is the wrap around from -PI to + PI

romulogcerqueira:
Hi @jmachowinski,

initially I created only two more operators, such as the existing ones (>, < and ==). 

However, after your question, I remembered I can use the current operators with the logical negative operator to achieve the same results.

Then I am closing this PR.

doudou:
> This looks a bit undefined.

Given that the strict `<` and `>` operators are already defined, these look like fairly natural extensions.

> Then I am closing this PR.

Don't. `!(a > b)` is a lot harder to read (ergo, makes the code much harder to maintain) than `a <= b`

romulogcerqueira:
> !(a > b) is a lot harder to read (ergo, makes the code much harder to maintain) than a <= b

@doudou, I agree with you. Should I reopen this PR?

jmachowinski:
I agree, as the < > was already defined, this is a natural extension. Sorry, didn't see this. 
Anyway, we should perhaps document the behaviour of these operations. 
It looks like we normalize to -M_PI <-> M_PI.
This then results in 181 degrees  < 179 degrees.

romulogcerqueira:
> It looks like we normalize to -M_PI <-> M_PI.

Actually the canonization is in `-M_PI + epsilon` <-> `M_PI`.

> This then results in 181 degrees < 179 degrees.

Correct.

doudou:
> This then results in 181 degrees < 179 degrees.

Not really, since Angle cannot represent 181 degrees by definition ... but I'm always for more documentation.

PR #110, opened 2017-06-30
Changes in the behaviour of Timeout::timeLeft() to be consistent with documentation

alcantara09:
When timeout is initialized with a null time, it is considered to be a maximum timeout. So timeleft should return a maximum value instead of zero.

Besides, the initilization fromMicroseconds and fromMiliseconds in base::Time was unsigned and the internal storage from time is signed. To be consistent with the documentation and with the expected behaviour of Timeout::timeleft() (return negative time after timeout is elapsed), the arguments from these methods were also changed.

Added also unit tests to cover this changes


2maz:
Thx, looks fine. Could  you squash the commits?

PR #111, opened 2017-07-03
Adding angular accelration to the rigid body acceleration type

alexfneves:
There is no angular acceleration type for a rigid body, so I added it into this type.

Signed-off-by: Alex Fernandes Neves <alexfneves@gmail.com>

### DIFF
--- a/src/samples/RigidBodyAcceleration.hpp
+++ b/src/samples/RigidBodyAcceleration.hpp
@@ -9,11 +9,17 @@ namespace base { namespace samples {
     {
         base::Time time;
 
-        /** Acceleration in m/s, world fixed frame of reference (East-North-Up) */
+        /** Linear acceleration in m/s2, world fixed frame of reference (East-North-Up) */
         base::Vector3d acceleration;
-	/** Covariance matrix of the acceleration
+	/** Covariance matrix of the linear acceleration
 	 */
         base::Matrix3d cov_acceleration;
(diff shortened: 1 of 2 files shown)


src/samples/RigidBodyAcceleration.hpp
```diff
@@ -9,12 +9,18 @@ namespace base { namespace samples {
...
-	/** Covariance matrix of the acceleration
+	/** Covariance matrix of the linear acceleration
 	 */
         base::Matrix3d cov_acceleration;
 
+        /** Angular acceleration in rad/s2, world fixed frame of reference (East-North-Up) */
+        base::Vector3d angular_acceleration;
+	/** Covariance matrix of the angular acceleration
+	 */
+        base::Matrix3d angular_cov_acceleration;
```
joaobrittoneto: What about `cov_angular_acceleration` to match with [RBS](https://github.com/rock-core/base-types/blob/master/src/samples/RigidBodyState.hpp#L70)?

alexfneves: I did another commit for the `cov_angular_acceleration`.

joaobrittoneto: :+1:

2maz:
Could you squash the two commits. If there are no further objections, I will merge then.

PR #114, opened 2017-08-15
DistanceImage: compute the scene point after sanity checks

saarnold:
Also:
- forward time stamp to point cloud
- fixed indentation to be consistent
- extended unit tests

### DIFF
--- a/src/samples/DistanceImage.hpp
+++ b/src/samples/DistanceImage.hpp
@@ -79,16 +79,15 @@ namespace samples
 	template <class Scalar_>
 	bool getScenePoint( size_t x, size_t y, Eigen::Matrix<Scalar_,3,1>& point ) const
 	{
-	    point = Eigen::Matrix<Scalar_,3,1>( (x*scale_x)+center_x, (y*scale_y)+center_y, 1.0 );
-
 	    // check bounds. x and y are always positive
-	    if( (x >= width) || (y >= height) ) 
-            return false;
+	    if( (x >= width) || (y >= height) )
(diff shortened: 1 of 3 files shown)


src/samples/DistanceImage.hpp
```diff
@@ -79,16 +79,15 @@ namespace samples
 	template <class Scalar_>
 	bool getScenePoint( size_t x, size_t y, Eigen::Matrix<Scalar_,3,1>& point ) const
 	{
-	    point = Eigen::Matrix<Scalar_,3,1>( (x*scale_x)+center_x, (y*scale_y)+center_y, 1.0 );
```
doudou: This would typically be a case for a unit test. "Don't write the output arg if input is wrong" is a common bug.

saarnold: I did add unit tests for both methods.

doudou: I actually did miss the tests ... need more sleep or less work maybe...

In any case, I don't believe that you verify that the point arg is left as-is if the coordinates are wrong.

saarnold: > I actually did miss the tests ... need more sleep or less work maybe...

No you didn't, I used the incorrect tense.

> I don't believe that you verify that the point arg is left as-is if the coordinates are wrong.

I didn't check this on purpose, because I wasn't sure if we want to enforce it.
The unit tests verify now that the args are not modified and I fixed it in getImagePoint as well.

PR #115, opened 2017-10-04
define Spline::derive

doudou:


doudou:
PIng

src/Spline.cpp
```diff
@@ -1147,4 +1147,18 @@ SplineBase *SplineBase::getSubSpline(double start_t, double end_t) const
     return new SplineBase(getGeometricResolution(), newCurve);
 }
 
+void SplineBase::derive(unsigned int order, SplineBase& result) const
+{
+    if (!SplineBase::getSISLCurve())
+        throw std::invalid_argument("cannot derive a singleton");
```
2maz: I would suggest using a prefix in the exceptions such as: base::geometry::SplineBase::derive: ...

src/Spline.cpp
```diff
@@ -1147,4 +1147,18 @@ SplineBase *SplineBase::getSubSpline(double start_t, double end_t) const
...
+
+    SISLCurve* newCurve;
+    int stat;
+    s1720(const_cast<SISLCurve*>(getSISLCurve()), order, &newCurve, &stat);
+    if (stat < 0)
+        throw std::runtime_error("failed to derive the curve");
```
2maz: same here

src/Spline.cpp
doudou: I like that, and I'll do it. However, this is the typical things that *if you believe it's better*, it would be great to actually advocate doing it to everyone...

doudou:
Done

PR #116, opened 2017-10-11
Add acceleration/angular_acceleration to RigidBodyState to have a

dmronga:
complete representation.

### DIFF
--- a/bindings/ruby/lib/base/typelib/rigid_body_state.rb
+++ b/bindings/ruby/lib/base/typelib/rigid_body_state.rb
@@ -11,7 +11,11 @@ Typelib.specialize_model '/base/samples/RigidBodyState_m' do
             :velocity => v3,
             :cov_velocity => m3,
             :angular_velocity => v3,
-            :cov_angular_velocity => m3)
+            :cov_angular_velocity => m3,
+            :acceleration => v3,
+            :cov_acceleration => m3,
+            :angular_acceleration => v3,
+            :cov_angular_acceleration => m3)
     end
 
     def Invalid; invalid end
--- a/src/samples/RigidBodyState.cpp
+++ b/src/samples/RigidBodyState.cpp
@@ -77,6 +77,10 @@ void RigidBodyState::invalidate()
     invalidateVelocityCovariance();
     invalidateAngularVelocity();
     invalidateAngularVelocityCovariance();
+    invalidateAcceleration();
+    invalidateAccelerationCovariance();
+    invalidateAngularAcceleration();
+    invalidateAngularAccelerationCovariance();
 }
 
 void RigidBodyState::initUnknown()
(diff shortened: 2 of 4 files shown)


joaobrittoneto:
There is a [RigibBodyAcceleration](https://github.com/rock-core/base-types/blob/master/src/samples/RigidBodyAcceleration.hpp). Doesn't it suit your needs?

dmronga:
Unfortunately not. I need Cartesian position, velocity and acceleration as well as their rotational parts in one type. If you want to implement e.g. a Cartesian position controller that considers the robot dynamics you usually need all of them. Also, I would only consider the RigidBodyState type as "complete" if it also specifies the acceleration

dmronga:
Any further concerns on this PR? I think the changes are reasonable...

chhtz:
Wouldn't extending `RigidBodyState` invalidate all logs which have previously been recorded? (I don't know exactly how serialization is implemented in rock ...)

Regarding "completeness": For that I'd expect to have the full N times N covariance -- but I don't think that has too many applications (I also don't like the current incomplete covariance in RBS, but that is another topic, as well)

doudou:
> Wouldn't extending RigidBodyState invalidate all logs which have previously been recorded? (I don't know exactly how serialization is implemented in rock ...)

It would. There is a tool to upgrade logs that should handle this particular case, but I don't know whether/if anybody ever uses it regularly (maybe @saarnold ? I've seen you commit some converters to base/types a while back), so "it's probably either broken or full of corner cases". Converters for base/types are currently stored here: https://github.com/rock-core/base-orogen-types/blob/master/src/log_migration.rb.

If some of you have the necessary motivation, I would suggest to write a converter, apply this PR and try `rock-upgrade` on the logs to see the result.

Generally speaking, long-term log management is a sore point in Rock, and to be honest all the robotic frameworks. I've done some major work in the right direction (https://github.com/rock-core/tools-pocolog/pull/17, https://github.com/rock-core/tools-log_tools/pull/8, https://github.com/rock-core/tools-syskit-pocolog) but nothing that is production-ready so far. Whether it will be made production ready in the near future depends on our companie's priorities (which might align to that goal hopefully soon, but I wouldn't hold my breath).

saarnold:
If I remember it correctly it should not be necessary to write a conversion in this case.
The new members are initialized (with NaN's) in the constructor and since this is a valid case a deep cast from the old to the new type should be enough. This is what rock-convert does as default.
However I also didn't use it in a long time so it should be evaluated if it still works as expected.

chhtz:
I don't really like the idea of having to convert old log files -- especially for a type which is used in like 99.9% of all scenarios. Even if conversion works, this would break backwards compatibility (i.e., if you want to run a log through an older version of your software stack, you'd additionally need to keep the old versions of your logs).

Even though this is not ideal either, I'd prefer introducing a new type for that (yes, that would be yet another type for something essentially the same).

dmronga:
Hi, I would like to reanimate this discussion. If changing the RigidBodyState type is unwanted, I would propose to add the following types to the base-types

base::Twist
----------------
base::Vector3d linear
base::Vector3d angular

base::Acceleration
-------------------------
base::Vector3d linear
base::Vector3d angular

base::samples::CartesianState
---------------------------------------
base::Time time
base::Twist twist

doudou:
While I think it will just add more confusion, I don't see any other way forward.

So, lukewarmly supporting :P

I would prefer another name for `CartesianState` though. Maybe something closer to where it should be used (assuming we still push for RBS to be used for state estimation) ?

doudou:
Ping

dmronga:
Hi, sorry for the (huge) delay. I will close this PR and open another one for the new data types

PR #120, opened 2018-09-12
Changes made in project entern

arneboe:
base-types was forked in project entern. This PR contains all changes that have been made (that have not been merged, yet).

I am in the process of closing all entern related repositories and would like to push anything upstream that might be useful.

Please review and either merge or close this PR :)


doudou:
Well, from my side, @raulg was guessing right. I'm really not much in favor of adding an arbitrary attribute field to Trajectory. If you have a hard time representing velocity profiles (i.e. point turn), it's probably because the type itself is too limited. Having string attributes will just be a horrible hack.

The two other commits look just fine.

chhtz:
IIRC the attributes added by Dennis were essentially only used by him for his thesis. I don't think anything else relies on that. I'd be fine with reverting/not merging these.

doudou:
> IIRC the attributes added by Dennis were essentially only used by him for his thesis. I don't think anything else relies on that. I'd be fine with reverting/not merging these.

Just cherry-picked and pushed the "{ empty spline }" commit. The depth map point size has already had been cherry-picked.

PR #121, opened 2018-10-09
Make SISL optional

arneboe:
The mainfest defines sisl as optional but base-types does not compile without sisl. This PR fixes that issue.

This reduces the number of dependencies that we need to install when using base-types outside of rock. In our use case we are dependent on base types but not on Spline/Trajectory, thus we do not need the SISL dependency.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,6 +10,7 @@ if(SISL_FOUND)
                Trajectory.hpp)
   set(sisl_cpp Spline.cpp
                Trajectory.cpp)
+  include_directories(${SISL_INCLUDE_DIRS})
 else()
   message(WARNING "SISL not found. Compiling without spline and trajectory support!")
 endif()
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -1,5 +1,22 @@
+
+find_package(SISL)
```
doudou: I would rather that this is controlled by an explicit option (i.e. WITH_SISL), to avoid having people getting an unexpectedly crippled base/types.

arneboe: > Perhaps we could always install `Spline.hpp` and `Trajectory.hpp`, but have it generate an `#error`, if SISL was not found?

This shifts the error to the compile phase instead of the configuration phase where it should occur.

In general I am fine with an option but it conflicts with having `optional=1` in the manifest.

doudou: > In general I am fine with an option but it conflicts with having optional=1 in the manifest.

Not necessarily. Autoproj will have to set the flag if sisl is not excluded, which would be done with 

~~~
unless Autoproj.manifest.excluded?('external/sisl')

end
~~~

Within the package definition block [in rock.core's 02stable.autobuild](https://github.com/rock-core/package_set/blob/master/04stable.autobuild#L44)

Note that the `package_enabled?` call is actually deprecated since Autoproj v2. To facilitate the transition between v1 and v2, I disabled the deprecation warnings by default and then forgot I did :(

doudou:
Closing this now that #134 has been merged.

PR #122, opened 2018-10-30
Added toTimeValues to base::time

JanWehrmann:
returns the time as a std::vector<int> containing
microseconds, milliseconds, seconds, minutes, hours and days
in that order.

### DIFF
--- a/src/Time.cpp
+++ b/src/Time.cpp
@@ -88,6 +88,33 @@ timeval Time::toTimeval() const
     return tv;
 }
 
+std::vector<int> Time::toTimeValues() const
+{
+    int64_t microseconds = this->microseconds;
+
+    int days = microseconds / 86400000000;
+    microseconds -= days * 86400000000;
+    int hours = microseconds / 3600000000;
+    microseconds -= hours * 3600000000;
+    int minutes = microseconds / 60000000;
+    microseconds -= minutes * 60000000;
+    int seconds = microseconds / 1000000;
+    microseconds -= seconds * 1000000;
+    int milliseconds = microseconds / 1000;
+    microseconds -= milliseconds * 1000;
+
+    std::vector<int> timeValues;
+    timeValues.reserve(6);
+    timeValues.push_back(static_cast<int>(microseconds));
+    timeValues.push_back(milliseconds);
+    timeValues.push_back(seconds);
+    timeValues.push_back(minutes);
+    timeValues.push_back(hours);
+    timeValues.push_back(days);
+
+    return timeValues;
+}
+
 std::string Time::toString(Time::Resolution resolution, const std::string& mainFormat) const
 {
     struct timeval tv = toTimeval();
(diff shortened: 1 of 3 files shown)


test/test.cpp
```diff
@@ -508,6 +508,17 @@ BOOST_AUTO_TEST_CASE(time_fromString)
 
 }
 
+BOOST_AUTO_TEST_CASE(time_toTimeValues)
+{
+    base::Time time = base::Time::fromMicroseconds(90061001001);
```
doudou: Great that you added tests ....

But I believe it would be best to have different values for each field. This test for instance would not catch having reported the milliseconds as days or seconds as microseconds or ...

JanWehrmann: Good point. Will change that.

src/Time.cpp
```diff
@@ -88,6 +88,32 @@ timeval Time::toTimeval() const
     return tv;
 }
 
+std::vector<int> Time::toTimeValues() const
+{
+    int64_t microseconds = this->microseconds;
+
+    int days = microseconds / 86400000000;
+    microseconds -= days * 86400000000;
```
chhtz: This line (`days * 86400000000`) will overflow for `days>10`, which is undefined behavior (i.e., it might accidentally work, but it could result in bogus results).

chhtz: Hm, thinking about it again, 86400000000 is already greater than max-int32, so this should actually work as intended (I made a `/` vs `*` error when first looking at it).

src/Time.cpp
```diff
@@ -88,6 +88,32 @@ timeval Time::toTimeval() const
...
+    int hours = microseconds / 3600000000;
+    microseconds -= hours * 3600000000;
+    int minutes = microseconds / 60000000;
+    microseconds -= minutes * 60000000;
+    int seconds = microseconds / 1000000;
+    microseconds -= seconds * 1000000;
+    int milliseconds = microseconds / 1000;
+    microseconds -= milliseconds * 1000;
+
+    std::vector<int> timeValues;
```
chhtz: For (tiny) efficiency improvement: `timeValues.reserve(6);`
Alternatively, consider using `std::array<int,6>`, instead (if this library is allowed to use C++11).

JanWehrmann: I don't think it is @saarnold ?

saarnold: As far as I know base-types does not contain C++11 code.
But I'm not sure about the current policy regarding C++11 in rock core libraries. Maybe @doudou can share some light here.

JanWehrmann:
Thanks for the feedback so far! As far as I can see I have resolved all of the issues that were pointed out. Is there anything else that needs to be revised?

doudou:
I don't think there is anything amiss ... @saarnold, @chtz ?

chhtz:
No objections from my side.

saarnold:
Looks good

PR #124, opened 2018-12-13
clarify RigidBodyAcceleration's usage

doudou:
I'm trying to not repeat the fiasco of the convention for velocities in RBS (see #99 and #100).

The original documentation stated that the frame of reference should be the world frame, but
within the discussion about RBS, people started (rightly) pointing out reasons why the body
frame would make more sense.

So, change the documentation of the type to match this idea. RBA is not widely used, and is
relatively new (when compared to RBS), so I hope this change will be accepted.

### DIFF
--- a/src/samples/RigidBodyAcceleration.cpp
+++ b/src/samples/RigidBodyAcceleration.cpp
@@ -2,7 +2,12 @@
 
 namespace base { namespace samples {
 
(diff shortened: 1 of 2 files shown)


saarnold:
:+1:

doudou:
Two acks ... I'm merging this RIGHT NOW ;-)

PR #125, opened 2019-01-25
add velocity profile visualization option in trajectory visualization

alcantara09:


viz/TrajectoryVisualization.cpp
```diff
@@ -45,6 +47,11 @@ void TrajectoryVisualization::setColor(const base::Vector3d& color)
     setDirty();
 }
 
+void TrajectoryVisualization::setMaxVelocity(double max_vel)
+{
+    max_velocity = max_vel;
```
doudou: The OSG nodes won't be updated (this is done by `updateMainNode`). Moreover, `updateMainNode` is called from within the OSG thread, while this method is called from the Qt thread.

The best is to lock (it's less critical than it sounds for something like this, but it's better), with

~~~
boost::mutex::scoped_lock lockit(updateMutex);
~~~

The *required* thing is to call `setDirty()` after you change the parameter, to force vizkit3d to call `updateMainNode`.

viz/TrajectoryVisualization.hpp
```diff
@@ -25,6 +25,7 @@ class TrajectoryVisualization: public Vizkit3DPlugin<base::Vector3d>
         TrajectoryVisualization();
         ~TrajectoryVisualization();
         void setColor(const base::Vector3d& color); 
+        void setMaxVelocity(double max_vel);
```
doudou: Would be best to declare MaxVelocity as a Q_PROPERTY as well, so that it shows in the property browser.

PR #126, opened 2019-02-04
add SonarVisualization for single beam sonar

alcantara09:
Add visualization for Scanning Sonar visualization using the new data
Type. It has the same functionality of SonarBeamVisualization plus the
option to visualize only the current beam or the full scan


viz/PluginLoader.cpp
```diff
@@ -79,6 +81,10 @@ namespace vizkit3d {
 	    {
 	        plugin = new vizkit3d::SonarBeamVisualization();
 	    }
+	    else if (pluginName == "SonarVisualization")
+	    {
+			plugin = new vizkit3d::SonarVisualization();
```
doudou: Indentation

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,176 @@
...
+{
+    { 
+        //boost::mutex::scoped_lock lockit(this->updateMutex);
+        this->full_scan = full_scan; 
+        data.clear();
+        if(!full_scan)
+        {
+            data.resize(1);    
+        }
+        else
```
doudou: `else if`

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,176 @@
...
+        {
+            if(motor_step != 0)
+            {
+                this->data.resize(num_steps+1);
+            }
+        }
+        
+    }
+    emit propertyChanged("Full Scan");
```
doudou: `FullScan` without the space

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,176 @@
...
+}
+
+bool SonarVisualization::isFullScan()
+{
+    return full_scan;
+}
+
+void SonarVisualization::updateMainNode(osg::Node* node)
+{
+    if (new_sonar_scan)
```
doudou: Please us an early return in this case, instead of having the whole method in a `if () ...` block

~~~
if (!new_sonar_scan)
   return;
~~~

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+double SonarVisualization::getMotorStep()
+{
+    return motor_step;
+}
+
+void SonarVisualization::setFullScan(bool full_scan)
+{
+    { 
+        //boost::mutex::scoped_lock lockit(this->updateMutex);
```
doudou: Locking, or not locking ? That is the question ...

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,176 @@
...
+void SonarVisualization::updateDataIntern(const base::samples::RigidBodyState& data)
+{
+    last_orientation = data;
+}
+
+void SonarVisualization::setMotorStep(double motor_step)
+{
+    { 
+        boost::mutex::scoped_lock lockit(this->updateMutex);
+        if (motor_step == 0)
```
doudou: This could (and IMO should) be out of the mutex-protected scope...

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,176 @@
...
+        data.clear();
+        if(!full_scan)
+        {
+            data.resize(1);    
+        }
+        else
+        {
+            if(motor_step != 0)
+            {
+                this->data.resize(num_steps+1);
```
doudou: I'm not so happy with the interaction between setMotorStep and setFullScan. I have the impression that one cannot exist without the other, in which case maybe it would be better to simply control the scan/not scan through the motor step ? (Just an idea).

If you do want to keep the full scan flag, I would simply call it `setScanning`. The "Full scan" thing is a bit hard to comprehend.

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,176 @@
...
+{
+    if(motor_step == 0)
+        return;
+    double current_pos = 0;
+    if(full_scan)
+        current_pos = ((data.bearings[0].rad+M_PI)/(2.0*M_PI))*num_steps;
+
+    SonarHelper bm;
+    bm.sonar_sample = data;
+    bm.orientation = last_orientation;
```
doudou: The placement of the plugin should be done through the transformer, not some internal pose handling. Note that if you do want to manually place it, you can manage its frame within vizkit3d. My point is: don't transform internally to the plugin.

viz/SonarVisualization.cpp
alcantara09: When using scanning sonars, one must set the motor step, otherwise nothing works. IMO it should be set in the constructor or something like this. Previously, the motor step was calculated taking into account two samples, however this was not robust and we have this information already.

doudou:
Is that expected ?

image

alcantara09:
yes

doudou:
.... could you change the message to describe what should be expected ? I was put off by the fact that the beam is not centered. I would do it myself, but I can't since I don't know whether the colors matter (if they do, describe them).

saarnold:
Could the motor step angle not be computed by successive scans?
Currently the default value is zero and in case of single scans needs to be set by the user.

doudou:
> Currently the default value is zero and in case of single scans needs to be set by the user.

This is because the "compute with successive scans" method is not robust at all if for whatever reason the widget is not getting updated fast enough. We've had a lot of trouble with this setting in all the sonar visualizations.

saarnold:
> This is because the "compute with successive scans" method is not robust at all if for whatever reason the widget is not getting updated fast enough.

Maybe we could keep it as an option, deactivated by default. 

What about using the beam width as initial setting? That would probably lead to gabs or overlapping between beams but would at least correlate with the area the signal comes from.

doudou:
> What about using the beam width as initial setting? That would probably lead to gabs or overlapping between beams but would at least correlate with the area the signal comes from.

Well ... duh. It would not be only better, it is the right thing to do (and it's simpler). The width of the beams should definitely be ... the `beam_width`

viz/SonarVisualization.cpp
```diff
@@ -0,0 +1,174 @@
...
+    drawArrays = new osg::DrawArrays( osg::PrimitiveSet::TRIANGLES, 0, pointsOSG->size() );
+    pointGeom->addPrimitiveSet(drawArrays.get());
+    
+    osg::ref_ptr<osg::Geode> geode = new osg::Geode;
+    geode->addDrawable(pointGeom.get());
+    mainNode->addChild(geode);
+    return mainNode;
+}
+
+double SonarVisualization::angleDifference(double a,double b)
```
doudou: 1. You don't seem to be using this
2. If you need something like that, just use [base::Angle](https://github.com/rock-core/base-types/blob/master/src/Angle.hpp)

alcantara09:
@saarnold, I stopped to think about the objective of the sonar visualization for a while. The full scan mode generates a really hard to interpret visualization, given that the robot's movement moves its position and the only reading that is aligned with the world is the current one. In the SonarWidget visualization, the scan is always in relation to the origin of the sensor, what does not hold in the 3d visualization. 

Given that we have the SonarWidget, I suggest we keep only the single and multibeam visualization for the current beam and do not support full scan mode.  What do you think about it?

saarnold:
> Given that we have the SonarWidget, I suggest we keep only the single and multibeam visualization for the current beam and do not support full scan mode. What do you think about it?

I would be OK with that. The keep old data flag is another way to visualize a batch of successive samples.

viz/SonarVisualization.hpp
```diff
@@ -0,0 +1,61 @@
...
+{    
+    Q_OBJECT
+    Q_PROPERTY(bool FullScan READ isFullScan WRITE showFullScan)
+    
+    public:
+        SonarVisualization();
+    
+        Q_INVOKABLE void updateSonar( const base::samples::Sonar& sample )
+        { return updateData(sample); }
+        Q_INVOKABLE void updateOrientation( const base::samples::RigidBodyState& orientation )
```
saarnold: This method is obsolete right?

alcantara09: yes

PR #127, opened 2019-02-07
Allow to set size of points in DepthMapVisualization.

maltewi:
Additionally:
* adds test program for DepthMapVisualization
* fixes crash when show_slope is selected on a DepthMap with only one row


doudou:
Looks good to me ... Since noone objected since February ...

Sorry for the long delay, @maltewi

Issue #132, opened 2019-03-21
Build breaks on system with only Qt5


jakobs:
9cb2df6dac66ba61e9164f989d53289e8bf09c9d added a check for qt4, which breaks the build on systems that only use Qt5, and not vizkit. Could we just move that check into viz/CMakelists.txt?

PR #133, opened 2019-03-21
Fix building without qt4

jakobs:
qt4 is an optional requirement only needed for the optional viz part of the package.
Moving it into the viz directory fixes building on systems, which don't have qt4 installed.


jakobs:
Fixes #132 on my system. Not sure if it helps the original intent of the commit...

doudou:
Looks like a very reasonable fix to me ...

PR #134, opened 2019-04-04
Make SISL truely optional

2maz:
@saarnold 

should be applied in combination with: rock-core/tools-service_discovery#9
base-orogen-types also requires adaption which is WIP: rock-core/base-orogen-types#29

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,4 +32,15 @@ if (EXISTS ${CMAKE_CURRENT_BINARY_DIR}/test)
     endif()
 endif()
 
+# SISL as optional dependency
+option(USE_SISL "set to OFF to disable the usage of SISL." ON)
+if(USE_SISL)
+    rock_find_cmake(SISL)
+    if(NOT SISL_FOUND)
+        message(FATAL_ERROR "The usage of SISL is requested, but the library could not be found."
+            " Either make sure SISL is installed, or disable the usage via"
+            " -DUSE_SISL=OFF")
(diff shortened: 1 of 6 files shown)


doudou:
Please don't make it dependent on the presence or not of SISL, which will only cause pain for people that for whatever reason won't have SISL available but expect it to be there.

Just add a USE_SISL option that people can turn to OFF explicitely if they don't want SISL.

2maz:
Updated

planthaber:
> Please don't make it dependent on the presence or not of SISL, which will only cause pain for people that for whatever reason won't have SISL available but expect it to be there.

Isn't this the default way ./configure scripts and any other installation of software with optional dependencies work? 

Even in most CMake libs, when a lib is not found I get a warning and have to provide e.g. `SISL_PATH` manually when it is not found.

What about a `WITH_SISL` define in Trajectory.hpp header that removed the code and produces an error "base-types was compiled without SISL, base::Trajectory is not available" on compile time when the header is used?

2maz:
> What about a `WITH_SISL` define in Trajectory.hpp header that removed the code and produces an error "base-types was compiled without SISL, base::Trajectory is not available" on compile time when the header is used?

Regarding the discussion on base/orogen/types then SISL is rather a dependency for Rock, while it should still be optional in base/types to facilitate the external use.
Thus I agree with @doudou to have the option and thereby an explicit way of controlling the need for SISL, while having a error reported at configuration of base/types if SISL is missing.

PR #136, opened 2019-04-16
spline: remove wrong validity check in Spline::interpolate

doudou:
The number of parameters must be the number of "real" points,
not just the number of points. Since the only way we could get
this number would be to scan the whole type array, just remove
the check altogether.

This fixes creating splines with parameters and derivatives

### DIFF
--- a/test/test_Spline.cpp
+++ b/test/test_Spline.cpp
@@ -4,9 +4,10 @@
 #include <base/geometry/Spline.hpp>
 
 
-BOOST_AUTO_TEST_SUITE(Spline)
+BOOST_AUTO_TEST_SUITE(SplineTests)
 
-using base::geometry::Spline;
+using namespace std;
+using namespace base::geometry;
 
 BOOST_AUTO_TEST_CASE(test_interpolate_with_nonempty_curve_updates_the_start_and_end_param_properly)
 {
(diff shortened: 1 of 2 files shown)


test/test_Spline.cpp
```diff
@@ -32,14 +33,48 @@ BOOST_AUTO_TEST_CASE(test_interpolate_with_no_points_sets_the_start_and_end_to_z
     BOOST_CHECK_EQUAL(0, spline.getEndParam());
 }
 
-BOOST_AUTO_TEST_CASE(test_interpolate_with_mismatching_number_of_points_and_parameters_throws)
+BOOST_AUTO_TEST_CASE(test_interpolate_accepts_points_and_parameters)
 {
+    auto points = vector<double>{1.0, 2.0, 3.0};
```
2maz: you imply c++11 for testing - better add using rock_activate_cxx11 in test/CMakeLists.txt then !?

doudou: Tbh I thought C++11 was enabled in the package overall. We've had that discussion with others in a PR. I'll start by proposing that ...

And don't use `rock_activate_cxx11` ... use cmake's own standard-setting mechanism. See https://github.com/rock-core/base-cmake/blob/master/modules/Rock.cmake and https://www.rock-robotics.org/rock-and-syskit/libraries/cpp_libraries.html#cxx_standard

doudou: #137

PR #141, opened 2019-10-29
Fixed an overflow issue in base::Time::toTimeValues

JanWehrmann:
and added additional tests to cover its occurrences
as well as some edge cases.


doudou:
Thanks !

Issue #109, opened 2017-06-29
test for Time::fromString does not pass


doudou:
When running the tests for `base::Time::fromString`, I get the following error:

~~~
test.cpp(426): fatal error in "time_fromString": critical check formatNow.toMicroseconds() == 1339668306001001 failed [1339686306001001 != 1339668306001001]
~~~

It looks like a misplaced 8 (686 vs 668), but I really am not sure what the right value should be. Does it run for you @2maz ?

2maz:
Test runs fine on Ubuntu 14.04 and the expected value should be correct (checked also with https://currentmillis.com/). What platform are you running the test on?

doudou:
> What platform are you running the test on?

16:04. I would guess it's a time zone issue (BRT vs CET)

2maz:
Yap, that is probably it: looks like timezone handling is not fully implemented through strptime

https://www.gnu.org/software/libc/manual/html_node/Low_002dLevel-Time-String-Parsing.html
```
%Z

    The timezone name.

    Note: Currently, this is not fully implemented. The format is recognized, input is consumed but no field in tm is set.

PR #143, opened 2019-12-09
fix test suite

doudou:
It was not passing. base/types ... !

### DIFF
--- a/bindings/ruby/test/test_matrixx.rb
+++ b/bindings/ruby/test/test_matrixx.rb
@@ -13,7 +13,7 @@ class TC_Eigen_MatrixX < Minitest::Test
         v = Eigen::VectorX.new(10)
         assert_equal(v.size, 10)
     end
-    
+
     def test_set
         m = Eigen::MatrixX.new(2,2)
         m[0,0]=2.0
@@ -50,7 +50,7 @@ class TC_Eigen_MatrixX < Minitest::Test
         assert_equal(m.row(2).to_a, v.to_a)
         assert_equal(m.row(1).to_a, v2.to_a)
     end
-    
+
     def test_set_col_add
         m = Eigen::MatrixX.new(2,2)
         v = Eigen::VectorX.new(2)
@@ -82,7 +82,7 @@ class TC_Eigen_MatrixX < Minitest::Test
         assert_equal(m[1,0],2)
         assert_equal(m[1,1],3)
     end
-        
+
     def test_matrix_dump_load
         m = Eigen::MatrixX.new(9,7)
         l = 9*7
@@ -104,13 +104,14 @@ class TC_Eigen_MatrixX < Minitest::Test
     end
 
     def test_dotV
-        m = Eigen::MatrixX.new(4,4)
-        4.times { |i| m[i,i] = i + 1 }
+        m = Eigen::MatrixX.new(4, 4)
+        4.times { |i| m[i, i] = i + 1 }
         a = Eigen::VectorX.from_a([1, 2, 3, 4])
         b = m.dotV(a)
         expected = Eigen::VectorX.from_a([1, 4, 9, 16])
         assert_kind_of Eigen::VectorX, b
-        assert(expected.approx?(b))
+        assert expected.approx?(b, 1e-3),
+               "expected #{b} to be approximately equal to #{expected}"
     end
 
     def test_jacobisvd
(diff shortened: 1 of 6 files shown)


2maz:
Looks like it is still not passing

```
../base/types/test/test_Time.cpp(110): fatal error: in "Time/fromString": critical check formatNow.toMicroseconds() == expected_utc_us + 1001 has failed [1339668306001001 != 1339671906001001]
```

doudou:
> Looks like it is still not passing

Damn.

A major difference with the behavior on my machine is that `formatNow.toMicroseconds()` returns the UTC time on your machine, local time on mine.

Could you run `timedatectl` ? To see whether your system clock is set in local time or UTC time. Could be a major difference.

doudou:
OK, so ... I dug even a little bit more.

strptime does not change the time fields when given a time zone. %Z is ignored on libc. %z sets a non-standard field (tm_gmtoff). See [the source code](https://github.com/lattera/glibc/blob/master/time/strptime_l.c)

`tm_gmtoff` is documented [here](https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html)

mktime assumes its argument to be in local time. There is timegm which does the same for UTC, but then it's also non-standard.

To be honest, I don't see a robust way to implement this method without spending an incredible amount of time (and probably still having a few corner-cases). I personally move to deprecate it wholesale and flag it as "don't use, unexpected behaviors may happen in various time zones. If you need to save/load time, save the seconds-since-epoch value returned by toMilliseconds()".

2maz:
> Could you run `timedatectl` ? To see whether your system clock is set in local time or UTC time. Could be a major difference.

$timedatectl                                               [12:23:06]
                      Local time: Do 2019-12-12 12:23:20 CET
                  Universal time: Do 2019-12-12 11:23:20 UTC
                        RTC time: Do 2019-12-12 11:23:20
                       Time zone: Europe/Berlin (CET, +0100)
       System clock synchronized: yes
systemd-timesyncd.service active: yes
                 RTC in local TZ: no

What issues would you see if we rely on http://man7.org/linux/man-pages/man3/tzset.3.html which is used by 'date in coreutils' and use "timezone" to set the value to be UTC.

btw that is still #109

doudou:
OK, misread the timestamps.

The value returned by fromString is 10:05, the expected value is actually 11:05 (once corrected by local time). The test value from the string is 12:05. So, I'm guessing problem detecting summer time since you're currently in winter (+1) and 14/06 was obviously summer. The code assumes that mktime is going to detect summer time. It maybe does, but using the current date and not the date in the tm struct ?

Which leads me back to my previous comment:

> To be honest, I don't see a robust way to implement this method without spending an incredible amount of time (and probably still having a few corner-cases). I personally move to deprecate it wholesale and flag it as "don't use, unexpected behaviors may happen in various time zones. If you need to save/load time, save the seconds-since-epoch value returned by toMilliseconds()".

tzset() **initializes** the timezone variable, which allows to correct for the effect of mktime, which turns a local time ("local time" being in this case the test value of 12:05) into GMT time.

2maz:
```
If you need to save/load time, save the seconds-since-epoch value returned by toMilliseconds()".
```
If you are in control of the full process ok, but you still might what to read a formatted timestamp from a different source.

I see the issue in the assumption of base Time implying local time..
If we want to avoid it altogether we have to base all values on UTC, i.e. enforce toString to return UTC and fromString as well. I tried to sketch something along these lines in #144

doudou:
> I see the issue in the assumption of base Time implying local time..

Just for the record, since I already said that in #144, but the safest assumption **by far** is to assume that base::Time stores UTC. `base::Time::now()` returns system time, which is in most cases UTC.

doudou:
> If you are in control of the full process ok, but you still might what to read a formatted timestamp from a different source.

My conclusion is that if you do need to do that, use another language that has a decent time representation and parsing, and use that language to convert to epoch. C/C++ does not give us the tools to do so, and I'd rather not pull a library in base/types to do that - or spend any more time to do it. I think that having the test suite for the robust parts of base/types passing is of widely more importance.

In addition, given that nobody was hit by this situation so far, I'm guessing that these methods are not widely used - or are at least not used to interpret historical data. Hence my position of "let's drop it".

For the record (again :P), C++20 will provide the tools for clock conversion it seems.

2maz:
> In addition, given that nobody was hit by this situation so far, I'm guessing that these methods are not widely used - or are at least not used to interpret historical data. Hence my position of "let's drop it".

I am using it just and would prefer not to be forced to look for a replacement - so updated #144.

PR #144, opened 2019-12-13
Timezone

2maz:


src/Time.cpp
```diff
@@ -28,6 +30,14 @@ Time Time::now()
     return Time(static_cast<int64_t>(t.tv_sec) * UsecPerSec + t.tv_usec);
 }
 
+Time Time::nowUTC()
```
doudou: This is actually wrong in most cases. gettimeofday does not return the local time, it returns the system time. Which on most default Linux distribution setups is already UTC. I personally assume that base::Time == UTC (because not doing so would be simply too full of corner-cases to account for)

The "culprit" that makes base::Time look like it stores local time is toString which breaks the time down to local time (using ... `localtime`). A non-localtime `toString` would be `gmtime`.

`fromString` and `fromTimeValues` are the real problem since `timegm`, the opposite of `gmtime`, is also a GNU extension. There are code snippets that should be usable to implement our own, but tbh I don't know how robust they actually are (and given how this stuff is complicated, I'd guess "not very")

Also, `timezone` is a non-standard GNU extension. I'd rather avoid using it with things like `musl` becoming rather common in the embedded world.

Again, I don't think this is worth our time *just* to have a proper fromString/fromTimeValues.

test/test_Time.cpp
```diff
@@ -96,18 +97,54 @@ BOOST_AUTO_TEST_CASE(fromString)
...
+    );
+    base::Time formatNowUTC1 = base::Time::fromString(
+        "2012-06-14--12.05.06Z:001001",
+        base::Time::Microseconds,
+        "%Y-%m-%d--%H.%M.%S%Z",
+        true,
+        0
+    );
+
+    BOOST_REQUIRE_EQUAL(formatNowUTC0.toMilliseconds(), formatNowUTC1.toMilliseconds());
```
doudou: You don't actually test that the conversion is exact. You test that the two calls return the same value - which could be 100% a wrong value. Use a epoch conversion website (I listed one in the comments) and actually figure out what value base::Time#toMicroseconds should have, and then test against that.

test/test_Time.cpp
```diff
@@ -96,18 +97,54 @@ BOOST_AUTO_TEST_CASE(fromString)
         " vs. converted: " << tzConverted.toString());
     // End time zone check
 
+    base::Time formatNowUTC0 = base::Time::fromString(
+        "2012-06-14--12.05.06Z:001001Z+0 GMT",
```
doudou: The first 'Z' is already the time zone info. Not sure what format you mean to implement.

src/Time.cpp
```diff
@@ -279,7 +303,22 @@ Time Time::fromString(const std::string& stringTime, Time::Resolution resolution
     tm.tm_isdst = -1;
     time_t time = mktime(&tm);
 
-    return Time(static_cast<int64_t>(time)*UsecPerSec + static_cast<int64_t>(usecs));
+    long tz_offset = 0;
```
doudou: As far as I can see (just looked at the code), what follows forgets one very important problem: `mktime()` assumes `tm` is in the _current_ time zone and converts it in what it believes that is the _current_ UTC.

I don't think we should allow for anything else than UTC to be stored in the base::Time

A.k.a. the conversion if you'd like to go to UTC is:
~~~
1. strptime: time in "UTC - stringTZOffset"
2. mktime: "UTC - stringTZOffset + localTZOffset"
3. need to add stringTZOffset and _remove_ local offset
~~~

2maz:
You are obviously right regarding the issue being fromString/toString, so updated this branch to address the issue by accounting for offset to UTC information via '%z'.

src/Time.hpp
```diff
@@ -55,9 +56,16 @@ namespace base
...
-         *     below seconds resolution if requested by the resolution argument
+         *     below seconds resolution if requested by the resolution argument,
+         *     and additionally suffixed by '%z' (see documentation of strftime)
+         *     the current timezone's offset to UTC
          **/
-	std::string toString(base::Time::Resolution resolution = Microseconds, const std::string& mainFormat = "%Y%m%d-%H:%M:%S") const;
+	std::string toString(base::Time::Resolution resolution = Microseconds,
+                const std::string& mainFormat = base::Time::DEFAULT_FORMAT) const;
+
+        static int64_t getTimezoneOffset(time_t time);
```
doudou: These methods really really need a very detailed documentation.

src/Time.hpp
```diff
@@ -55,9 +56,16 @@ namespace base
...
          * \param resolution Resolution which the string should present
          * \param mainFormat Main format to use -- this is passed to strftime and appended by ':' plus the
-         *     below seconds resolution if requested by the resolution argument
+         *     below seconds resolution if requested by the resolution argument,
+         *     and additionally suffixed by '%z' (see documentation of strftime)
+         *     the current timezone's offset to UTC
          **/
-	std::string toString(base::Time::Resolution resolution = Microseconds, const std::string& mainFormat = "%Y%m%d-%H:%M:%S") const;
+	std::string toString(base::Time::Resolution resolution = Microseconds,
+                const std::string& mainFormat = base::Time::DEFAULT_FORMAT) const;
```
doudou: This whole block's indentation is broken on GH's PR view, I'm guessing there are tabs.

src/Time.cpp
```diff
@@ -272,16 +291,66 @@ Time Time::fromString(const std::string& stringTime, Time::Resolution resolution
...
-    // (http://pubs.opengroup.org/onlinepubs/007904975/functions/strptime.html)
 
-    tm.tm_isdst = -1;
-    time_t time = mktime(&tm);
+    time_t time;
+    if(tzInfo.empty())
+    {
+        tm.tm_isdst = -1;
+        time = mktime(&tm);
+    } else {
```
doudou: Consistency in `{` placement. This looks bizarre.

src/Time.cpp
```diff
@@ -272,16 +291,66 @@ Time Time::fromString(const std::string& stringTime, Time::Resolution resolution
...
+        // full offset to UTC
+        int64_t tzOffset = Time::tzInfoToSeconds(tzInfo);
+        //// now we are in utc input
```
doudou: given the very next comment, we're really not.

src/Time.cpp
```diff
@@ -272,16 +291,66 @@ Time Time::fromString(const std::string& stringTime, Time::Resolution resolution
...
+        tm.tm_sec += tzOffset;
+
+        // mktime interpretes tm as local time
+        // We have remove the local timezone's offset (at the selected date) to UTC, to really get UTC
+        tm.tm_isdst = -1;
+        time = mktime(&tm);
```
doudou: This whole thing is very dubious actually.

The time offset does not define the existence of DST, only the time zone name can. Algiers and Paris have the same base offset, Paris had DST in 2019, Algiers did not. 

I guess that one workable solution is to assume that the time-offset marker already includes the DST (it probably does, because of the above comment) and not handle DST at all.

src/Time.hpp
```diff
@@ -113,22 +121,29 @@ namespace base
...
-        *                   i.e. the string is always interpreted as a local time.
+        * \param mainFormat valid main format for strptime, e.g."%Y%m%d-%H:%M:%S". Note
+        *                   that the timezone field %Z is parsed but basically
+        *                   ignored.
         *                   The default format converts the value returned by
         *                   fromString (i.e. Time::fromString(time.toString()) is
         *                   the identity).
+        * \returns a Time object representing UTC if the input string is
+        *           suffixed by the offset to UTC '%z', e.g. +0230 for two hours
+        *           and 30 minutes (east of GMT), and localtime otherwise
```
doudou: not 'localtime', 'the time represented by the string'. Which could itself be UTC.

src/Time.cpp
2maz: The time offset derives from '%z' strftime:
```
The +hhmm or -hhmm numeric timezone (that is, the hour and
              minute offset from UTC).
```
so DST is included.

A requirement of the implementation is, however, that setting of tm_isdst here has to be mirrored within getTimezoneOffset.
If we want getTimezoneOffset to correspond to 'timezone' for the current time, then this needs to be -1, otherwise we can fix it to 0.

src/Time.hpp
2maz: IMO "the time represented by the string" misses out that the string is interpreted for the current timezone. If the current timezone is GMT then localtime == UTC?

src/Time.cpp
doudou:
~~~
// now we move to the local timezone
~~~

Waitaminute.

Let's get back one step: do you want the Time as returned by fromString to be represented in the local time zone ?

src/Time.hpp
```diff
@@ -52,68 +53,113 @@ namespace base
          */
         std::vector<int> toTimeValues() const;
 
-	/** Convert time into a string
+        /** Convert time into a string
```
doudou: `to a string representing the local time` ?

src/Time.cpp
```diff
@@ -259,18 +284,72 @@ Time Time::fromString(const std::string& stringTime, Time::Resolution resolution
...
+
+        // Identify offset to UTC in the current timezone at 'time'
+        int64_t localTimezoneOffset = Time::getTimezoneOffset(time);
+        // now we move to the local timezone
+        time -= localTimezoneOffset;
```
doudou: I'm going to assume that the documentation you wrote for fromString is what you actually intend to do, which is

~~~
        * \returns a Time object representing UTC if the input string is
        *           suffixed by the offset to UTC '\%z', e.g. +0230 for two hours
        *           and 30 minutes (east of GMT), and localtime otherwise
~~~

If the intention is indeed to return a base::Time in UTC, then the only thing needed is to apply the offset contained in the string. getTimezoneOffset is not needed.

src/Time.cpp
doudou: OK. I'm sorry if you are receiving notifications for all the comments I've left and then removed. This is as I expected, mind-numbing. Anyways. I misread the whole thing, my mind is getting tired of juggling with time zones.

I think this needs a lot more comments

~~~
else
{
  // We have to go through a few intermediate "time zones" to reach UTC
  // The problem we try to work around here is that mktime converts a broken-down time
  // in the local time zone to UTC, but what we have is a time in UTC and want to convert
  // it to UTC

  // 'tm' is the time represented by the string. It is `tzInfo` seconds away from UTC
  tm.tm_sec += tzOffset;
  // OK, now `tm` is the string's time in UTC. Note that mktime handles wraparounds
  tm.tm_isdst = -1;
  time = mktime(&tm);
  // Since mktime expected `tm` to be in local time, `time` is UTC with the local time
  // zone offset as it was at `tm` applied in opposite (e.g. time would be UTC-3 if
  // executed on a machine whose current time zone is UTC+3).
  // We have to correct with the local offset to finally get UTC
  int64_t localTimezoneOffset = Time::getTimezoneOffset(time);
  time -= localTimezoneOffset;
}
~~~

src/Time.hpp
doudou: Yes, you are right. Which means that the Time object represents ... whatever.

That is "the time from the string offset with the current time zone offset".

Maybe a better description would be 

~~~
If the input string contains a time offset representation of the form +HHMM (+0230 for 2:30 east of UTC) and the format captures it with `%z`, the value returned is the UTC time. Otherwise, the string is interpreted as a time in the local time zone and converted to UTC accordingly.
~~~

src/Time.cpp
2maz: Indeed mind-numbing, but glad you are still up for reviewing. Updated according to your suggestions here.

src/Time.hpp
2maz: Updated with slight changes to point out that right now only the +HHMM at the end of the string is supported.

doudou:
Hey. Looks for me now.

I suggest we first merge the other PR without the time-related changes, and then this one. This is a big piece of work, I think it'd be best to keep a separate merge/PR commit for it.

If you agree, I can do the rebasing mumbo-jumbo to clean this up.

2maz:
> I suggest we first merge the other PR without the time-related changes, and then this one. This is a big piece of work, I think it'd be best to keep a separate merge/PR commit for it.

Agreed

> If you agree, I can do the rebasing mumbo-jumbo to clean this up.

Sure, thanks taking care of this.

doudou:
OK. I'll get those ready, but wait for the bundler mess fix to get them through CI before I actually merge them.

The fix got merged and the developer who fixed it said he'll try and get a point release out quickly... Fingers crossed.

PR #146, opened 2020-05-04
RigidBodyStateSE3

dmronga:
This PR adds the data type RigidBodyStateSE3 which includes pose, twist, spatial acceleration and wrench in one structure, as well as some missing subtypes. The motivation behind this is to have a complete representation of the state of a rigid body for dynamic control applications. The necessity to have a new type for this has been discussed in  PR #116. 

Two words on the data type:
- We opted for the name RigidBodyStateSE3 (SE3 stands for Special Euclidean Group with dimension 3) because we would like to push for group theoretic data types. The advantage is that operations on these types are clearly defined (for example only multiplication between two SE3 elements makes sense). Also most of modern robotics is built on the mathematical foundations of group theory and differential geometry, see e.g. [here](http://hades.mech.northwestern.edu/images/7/7f/MR.pdf) as reference
- The type can be seen in analogy to base/JointState where we have position, speed, acceleration and effort (force/torque)

@shivesh1210


src/samples/RigidBodyStatesSE3.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include <base/NamedVector.hpp>
+#include <base/Time.hpp>
+#include <base/RigidBodyStateSE3.hpp>
+
+namespace base { namespace samples {
+
+/** Named vector of base::RigidBodyStateSE3 with sampled Time. */
+struct RigidBodyStatesSE3 : public base::NamedVector< base::RigidBodyStateSE3>
+{
```
doudou, line 12: That ... won't do. The `s` is way too subtle.

shivesh1210: I agree that we should have extensive unit testing for this type and also very clear documentation about the notation and links to other notations in the literature. There are many notations in the literature for e.g. Featherstone, Lynch & Park, Murray & Sastry etc and there are subtle differences which can create big confusions across components which would use this type in future. 

Regarding the naming, I would insist that we have SE3 in the naming unless our RigidBodyState type is so generic that we just specify which group the rigid body "lives" in and it provides all the necessary group operations and coordinate transformations. Some examples of rigid body groups along with their operations are (R2,+), (R3,+), (SE2, *), (SE3, *), (SO2, *), (SO3, *) etc. All these groups are subspaces of SE3 and hence a SE3 type can capture them but it is often an overkill. In future, we should work towards general group theoretic types in RoCK but for the moment SE3 would be fine.

doudou: Again, I won't be fighting over the name. I'd just like you to understand my perspective.

Think about it this way: you are doing a great work for people that don't give a sh*t about group-theoretic whatever (I'm not one of them). You're doing a great work because by defining this type cleanly, you are making sure it will "just work" sensibly.

Naming your type this way won't help understand why they should be using it. Finding a better name might. Then, please do explain in great lengths why this type is internally consistent thanks to its math roots. In the docs.

At the end of things, I like the mathematical cleanliness of the approach *not* because it's group-theoretic whatever, but because I trust that someone that would have done its homework like you obviously did will create something that will work sensibly (hence, avoiding problems). Having SE3 in the name ? Not helping me understand why I should be.

I can't be using this type for state estimation (lacking essential info). Does SE3 in the name help me understand that ?

If this is essentially meant for control or motion planning, maybe BodyMotionState (not liking that, just trying to start a conversation here) ?

shivesh1210: I think our chosen naming convention "RigidBodyStateSE3" satisfies the needs of both a novice robotics developer (RigidBodyState tells this type describes the position, velocity, acceleration and wrench state of a rigid body) and advanced user (SE3 tells that this rigid body motion is described in the Special Euclidean Group of dimension 3). If somebody does not know what the latter part means, they can look it up and they will only learn something useful. Understanding geometry in robotics is of prime importance, please take a look here: http://rss2017.lids.mit.edu/docs/invitedtalks/park-manifoldsgeometryandrobotics.pdf.

doudou: Again, you don't have to convince me of the underlying approach ...

And I don't want to have a long drawn fight over this, so let's stick to RigidBodyStateSE3

The two remaining issues are:
- `RigidBodyStatesSE3`. Anything that does not involve parsing a lower case `s` appearing in the middle of a huge identifier would be better.
- unit tests

doudou: And since we're here sharing references, this one's also interesting if you haven't read it yet: [Part 1](https://ieeexplore.ieee.org/abstract/document/6476696) and [Part 2](https://ieeexplore.ieee.org/document/6524117)

dmronga: I will take care of the unit tests
How about RigidBodyStateSE3Vector for the Vector type?

dmronga: I added some unit tests and updated the documentation. Minor changes have also been made to base/samples/Wrench and base/Wrench to align it with the analogue Twist/Acceleration types. 
@shivesh1210 Can you double check the docs?

shivesh1210: It looks fine to me. It can be merged from my perspective.

doudou:
Looks good to me. Given that you were the one recently introducing Twist and Wrench, I'd even be OK with changing the type definitions (that's why we need official releases ... to have stricter rules on that)

@2maz ... ?

doudou:
> Looks good to me. Given that you were the one recently introducing Twist and Wrench, I'd even be OK with changing the type definitions (that's why we need official releases ... to have stricter rules on that)
> 
> @2maz ... ?

Sorry, disregard the second part of the message. Misread the github diff (it was showing only the latest changes ...)

src/samples/Acceleration.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef BASE_SAMPLES_ACCELERATION_HPP
+#define BASE_SAMPLES_ACCELERATION_HPP
+
+#include <base/Acceleration.hpp>
+#include <base/Time.hpp>
+
+namespace base { namespace samples {
+
+/** Spatial Acceleration with sampled time and frame ID*/
+struct Acceleration : public base::Acceleration{
```
2maz, line 10: style: the main style in this package uses brace on next line

src/samples/Twist.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef BASE_SAMPLES_TWIST_HPP
+#define BASE_SAMPLES_TWIST_HPP
+
+#include <base/Time.hpp>
+#include <base/Twist.hpp>
+
+namespace base { namespace samples {
+
+/** Twist with sampled time and frame ID*/
+struct Twist : public base::Twist{
```
2maz, line 10: style: see prev comment

test/test_RigidBodyStateSE3.cpp
```diff
@@ -0,0 +1,121 @@
+#include <boost/test/unit_test.hpp>
+#include <base/samples/RigidBodyStateSE3.hpp>
+
+using namespace std;
+using namespace base;
+
+BOOST_AUTO_TEST_SUITE(RigidBodyStateSE3Tests)
+
+BOOST_AUTO_TEST_CASE(rbs_se3_initialization){
```
2maz, line 9: style: see prev comment

2maz:
@dmronga thanks for taking the effort

PR #149, opened 2020-06-05
ruby: fix spurious test failures

doudou:
Two tests were not zeroing-out their matrix/vector while it was needed. Add the definition of `#zero` and `.Zero` for all types where it makes sense, and fix the tests.


doudou:
Ping @g-arjones

PR #153, opened 2021-01-26
Rbs se3 viz

mmaurus:
Added vizkit3d plugin for Wrenches and new RigidBodyStateSE3


viz/PluginLoader.cpp
```diff
@@ -39,6 +41,8 @@ namespace vizkit3d {
 	    pluginNames->push_back("SonarVisualization");
 	    pluginNames->push_back("DepthMapVisualization");
 	    pluginNames->push_back("DistanceImageVisualization");
+		pluginNames->push_back("RigidBodyStateSE3Visualization");
+		pluginNames->push_back("WrenchVisualization");
```
doudou, line 45: Wrong indentation (happens elswhere in the patch).

Rock style guide is 4 space indent, no tabs. I'm guessing that you are using tabs.

viz/WrenchModel.cpp
```diff
@@ -0,0 +1,147 @@
...
+    osg::ref_ptr<osg::PositionAttitudeTransform> arr = primitivesfactory->createCircularArrow(0.1, 0.02, 12, 32, 1.5*M_PI, 2.0, color);
+
+    return arr;
+}
+
+
+void WrenchModel::buildGeometry() {
+
+    
+
```
doudou, line 30: Please avoid unecessary leading/trailing lines.

viz/WrenchModel.cpp
```diff
@@ -0,0 +1,147 @@
...
+    else {
+        if (wrench.force.isZero()) {
+            // if force is zero vector, disable
+            setChildValue(force_node, false);
+        } 
+        else {
+            // update transform
+            osg::Vec3d force(wrench.force.x(), wrench.force.y(), wrench.force.z());
+            Q.makeRotate(osg::Vec3d(0,0,1), force);
+            //force_node->setScale(osg::Vec3f(resolution, resolution, resolution*force.length()));
```
doudou, line 94: Don't commit commented-out code. If it is commented out, it's not part of the implementation and just adds noise.

doudou:
Not a requirement, but we've started to create a while ago "interactive" unit tests for visualization, that are in `test/viz/`. They basically work by showing some data and asking the user/tester whether some stuff appear as expected.

Would you consider creating those for these two visualizers ? It would really help.

viz/CMakeLists.txt
```diff
@@ -37,8 +40,11 @@ rock_vizkit_plugin(base-viz
...
         PointcloudVisualization.hpp
         DepthMapVisualization.hpp
+        RigidBodyStateSE3Visualization.hpp
+        WrenchModel.hpp
+        WrenchVisualization.hpp
         ${OPTIONAL_HPP}
     DEPS base-types
     LIBS ${Boost_SYSTEM_LIBRARY}
-    DEPS_PKGCONFIG base-logging
+    DEPS_PKGCONFIG base-logging PrimitivesFactory osgViz
```
g-arjones, line 49: I really believe we should consider other alternatives before adding `osgViz` as a "hard" dependency in `base/types`.

This is also breaking rock-core's build. 

@doudou

doudou: In practice, base/types already depends on `gui/vizkit3d` which depends on osgViz. The (optional) dependency should definitely be explicitly added to the manifest, though.

g-arjones: Yes, that will be a problem in constrained environments (i.e embedded computers running iodrivers_base based drivers). Can we please don't do that?

doudou: The keyword was **optional** ;-). The dependency on osgViz that this PR adds is also optional as the whole `viz/` folder is disabled by the Rock CMake macros if vizkit3d is not available or if [ROCK_VIZ_ENABLED](https://github.com/rock-core/base-cmake/blob/master/modules/Rock.cmake#L217) is set to OFF.

base/types already depends **optionally** on `gui/vizkit3d`. In constrained environments, you usually exclude `gui/.*` which removes that particular dependency

g-arjones: > the whole viz/ folder is disabled by the Rock CMake macros if vizkit3d is not available or if ROCK_VIZ_ENABLED is set to OFF

I missed that, sorry. I guess I was misled by the breaking but (which is probably breaking for some other reason I didn't look into). Did this pass on your CI?

g-arjones:
```
-- Checking for module 'PrimitivesFactory'
--   No package 'PrimitivesFactory' found
CMake Error at /usr/share/cmake-3.10/Modules/FindPkgConfig.cmake:415 (message):
  A required package was not found
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPkgConfig.cmake:593 (_pkg_check_modules_internal)
  /buildbot/squidbot-build/build/install/base/cmake/share/rock/cmake/Rock.cmake:288 (pkg_check_modules)
  /buildbot/squidbot-build/build/install/base/cmake/share/rock/cmake/Rock.cmake:383 (rock_find_pkgconfig)
  /buildbot/squidbot-build/build/install/base/cmake/share/rock/cmake/Rock.cmake:649 (rock_target_definition)
  /buildbot/squidbot-build/build/install/base/cmake/share/rock/cmake/Rock.cmake:811 (rock_library_common)
  viz/CMakeLists.txt:11 (rock_vizkit_plugin)
```

doudou: I don't know what PrimitivesFactory is or where it comes from. I was looking at osgViz since it was what you mentioned in your comment.

Didn't check CI yet.

mmaurus: PrimitivesFactory is a module in osgViz

PR #155, opened 2021-07-26
add base::Time::monotonic()

doudou:
Meant for timeout calculations.

### DIFF
--- a/src/Time.cpp
+++ b/src/Time.cpp
@@ -9,6 +9,9 @@
 #include <time.h>
 #include <limits>
 #include <regex>
+#include <chrono>
+
+using namespace std;
 
 namespace base {
 
@@ -29,6 +32,16 @@ Time Time::now()
     timeval t;
     gettimeofday(&t, 0);
     return Time(static_cast<int64_t>(t.tv_sec) * UsecPerSec + t.tv_usec);
+}
+
+Time Time::monotonic()
+{
+    // Note: C++11 statics are thread safe
+    static auto monotonicClock = std::chrono::steady_clock();
+
+    auto tp = monotonicClock.now().time_since_epoch();
+    auto us = chrono::duration_cast<std::chrono::microseconds>(tp);
+    return Time(us.count());
 }
 
 bool Time::operator<(const Time& ts) const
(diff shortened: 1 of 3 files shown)


src/Time.cpp
```diff
@@ -31,6 +34,14 @@ Time Time::now()
     return Time(static_cast<int64_t>(t.tv_sec) * UsecPerSec + t.tv_usec);
 }
 
+static auto monotonicClock = std::chrono::steady_clock();
```
2maz, line 37: Why don't you keep monotonicClock inside the function scope of Time::monotonic? It is not used anywhere else.

doudou: Because then initialization would not be thread safe (clock.now() is thread safe in C++11)

2maz: According to the standard 11 draft: http://www.open-std.org/JTC1/SC22/WG21/docs/papers/2011/n3242.pdf
6.7.4:

"If control enters the declaration concurrently while the variable is being initialized, the concurrent execution shall wait for
completion of the initialization."

(see also https://iamroman.org/blog/2017/04/cpp11-static-init/)

Related to this for gcc: 
```
 -fno-threadsafe-statics
           Do not emit the extra code to use the routines specified in the C++
           ABI for thread-safe initialization of local statics.  You can use
           this option to reduce code size slightly in code that doesn't need
           to be thread-safe.

doudou: Cool. Did not know that C++11 fixed that. Will change.

doudou: Fixed.

2maz (approved):
I suggest to remove the using statement in L14 and just stick to std::chrono in L43 - for the rest you used std:: anyhow.

doudou:
I would rather add another commit to remove all the `std::` stuff ...

2maz:
I leave the choice to you, and at the same time add another option, how about adding:
using std::chrono within the scope of Time::monotonic?

doudou:
I meant "for the whole file". See 9077b5e.

2maz:
> I meant "for the whole file". See [9077b5e](https://github.com/rock-core/base-types/commit/9077b5ec00b477bea77483a5a7367e3934491d4d).

I feared that ;) - but your choice, so accepted (review dialogue does not show for unknown reasons)

PR #160, opened 2022-08-09
fix(ruby): return value of SplineBase#simplify

doudou:
rice would generate an error when trying to get the 'vector'. I think it should work, but had other things to solve .. so I created an intermediate method does the conversion manually instead.


jhonasiv (approved):

caioaamaral (approved):
I didn't quite get what was broken before

doudou:
rice would generate an error when trying to get the 'vector<double>'. I think it should work, but had other things to solve .. so I created an intermediate method does the conversion manually instead.

PR #165, opened 2023-09-18
Automatically determine size of jpeg frames

chhtz:
Sometimes, when reading a jpeg file you don't want to fully decompress it, but keep it as compressed imaged. `frame::Frame` still wants to know the size of the image, which can be determined using libjpeg.

Open points: 
* Shall this functionality be moved elsewhere? (e.g., `FrameHelper`)
* Additional functionality (maybe also in `FrameHelper`, such as `loadImageCompressed` which keeps the jpeg images in compressed mode)
* No specific error handling yet
* No unit test yet (I tested this with the `FrameHelper` test_suite, but did not make a PR for that)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -114,6 +114,7 @@ rock_library(
         samples/Event.hpp
         samples/EventArray.hpp
         ${OPTIONAL_HPP}
+    DEPS_CMAKE JPEG
     DEPS_PKGCONFIG 
         base-logging
         eigen3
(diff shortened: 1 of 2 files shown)


doudou:
I think this should be moved to FrameHelper as a "readJPEG" method that returns the frame.

chhtz:
I added corresponding functions to FrameHelper (added a helper function to JpegConversion):
https://github.com/rock-core/perception-jpeg_conversion/pull/3
https://github.com/rock-core/perception-frame_helper/pull/18

chhtz:
Closing this, as the feature has been merged into FrameHelper (see links above).

PR #167, opened 2024-05-16
Add Logmessage type

planthaber:
Adds a LogMessage type (a timestamped string)

Provides ability to:
* put text-based markers into binary logs, like when an experiement started 
* use a port to show/record the time of state changes
* prepeares for an (to be published) orogen plugin that copies all stdout/stderr output to an output port to be able to print all output with an application rather than searching in the text log files

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -62,6 +62,7 @@ rock_library(
         JointsTrajectory.hpp
         JointTransform.hpp
         Logging.hpp
+        LogMessage.hpp
         Matrix.hpp
         NamedVector.hpp
         Point.hpp
@@ -114,7 +115,7 @@ rock_library(
         samples/Event.hpp
         samples/EventArray.hpp
         ${OPTIONAL_HPP}
-    DEPS_PKGCONFIG 
+    DEPS_PKGCONFIG
         base-logging
         eigen3
 )
--- a/src/LogMessage.hpp
+++ b/src/LogMessage.hpp
@@ -1,0 +1,24 @@
+#ifndef LOG_MESSAGE_HPP
+#define LOG_MESSAGE_HPP
+
+#include <string>
+#include <base/Time.hpp>
+
+namespace base
+{
+
+    /**
+     * @brief Log message type to be able to e.g. write timestamped messages/markers into log files
+     * 
+     */
+    class LogMessage
+    {
+        public:
+            base::Time timestamp;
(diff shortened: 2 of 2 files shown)


planthaber:
Here is the PR for the orogen part: https://github.com/rock-core/base-orogen-types/pull/41

chhtz:
I suggest using `time` instead of `timestamp` to be compatible with other samples and related tooling (in case there ever is a use case for that). 
And maybe simply `base::TimeStamped<std::string>` would work as well? (I haven't actually tried this)

planthaber:
> I suggest using `time` instead of `timestamp` to be compatible with other samples and related tooling (in case there ever is a use case for that). And maybe simply `base::TimeStamped<std::string>` would work as well? (I haven't actually tried this)

the TimeStamped template dioes not work for container types with gccxml: 
> #<Thread:0x000057f977ce2200 /opt/workspace/install/lib/ruby/2.7.0/typelib/gccxml.rb:1043 run> terminated with exception (report_on_exception is true):
    /opt/workspace/install/lib/ruby/2.7.0/typelib/gccxml.rb:656:in `block (2 levels) in resolve_type_definition': undefined method `each_field' for #<Typelib::ContainerType: /std/string> (NoMethodError)

But changed the time filed

doudou:
> use a port to show/record the time of state changes

I would rather suggest we modify orogen to have a timestamped state change that uses the state enum. string is not hard real time, you can't make it mandatory at low level.

planthaber:
> I would rather suggest we modify orogen to have a timestamped state change that uses the state enum. string is not hard real time, you can't make it mandatory at low level.

Yes, but that was just an example why the type may be useful in general. The original reason I need the type is the text logging orogen plugin I mentioned (which is only needed for debugging anyways). 

It already works, when the orogen plugin is active, stdout and stderr are split up into writing to the log file (as it is now) and additionally writing the content LogMessage to a port. This makes it possible to collect and display all produced text in a single application rather than searching though the log files in the bundle (including filtering on task/context level). But here is where we need the timestamp in order to at least have the possibility to stream align the messages.

Also, when the strings are set in the constructor or confiugure hook, it should be fine with the real-time contraint right? Then it's on the implementation side, and it is not mandatory at all.

On the plugin this is probably not real-time, but it is for debugging purposes only and can be deactivated when needed, it will be inactive when no output is written to stdout, anyways.

PR #171, opened 2024-12-30
feat: Implement a method to saturate joints according to their limits

DeboraTahara:


DeboraTahara:
@doudou ping

doudou (approved):

src/JointLimits.cpp
```diff
@@ -33,4 +33,27 @@ void JointLimits::validate(const samples::Joints& joints) const
...
+{
+    samples::Joints new_joints;
+    new_joints.resize(joints.size());
+    new_joints = joints;
+    bool saturated = false;
+    if (joints.hasNames()) {
+        for (size_t i = 0; i < joints.size(); i++)
+        {
+            auto result = (*this)[joints.names[i]].saturate(joints[i]);
+            saturated = saturated ? saturated : result.first;
```
jhonasiv, line 46:
```suggestion
            saturated ||= result.first;
```

src/JointLimits.cpp
```diff
@@ -33,4 +33,27 @@ void JointLimits::validate(const samples::Joints& joints) const
...
+            new_joints[i] = result.second;
+        }
+    }
+    else {
+        for (size_t i = 0; i < joints.size(); i++)
+        {
+            auto result = (*this)[i].saturate(joints[i]);
+            saturated = saturated ? saturated : result.first;
```
jhonasiv, line 54: same thing

src/JointLimits.cpp
DeboraTahara: It's not the same thing!

chhtz: There is no `||=`  operator, but with `|=` these are equivalent: https://godbolt.org/z/73cM6coKr

jhonasiv (approved):

PR #172, opened 2025-01-30
feat: allow to initialize a frame without filling its image buffer

doudou:


src/samples/Frame.cpp
```diff
@@ -26,13 +26,13 @@ bool frame::frame_size_t::operator!=(const frame::frame_size_t& other) const
...
 
-frame::Frame::Frame() : image(), size(), data_depth(0), pixel_size(0), frame_mode()
+frame::Frame::Frame() : image(), size(), data_depth(0), pixel_size(0), frame_mode(MODE_UNDEFINED)
 {
     setDataDepth(0);
     reset();
 }
 
-frame::Frame::Frame(uint16_t width, uint16_t height, uint8_t depth, frame::frame_mode_t mode, const uint8_t val, size_t sizeInBytes)
+frame::Frame::Frame(uint16_t width, uint16_t height, uint8_t depth, frame::frame_mode_t mode, int16_t val, size_t sizeInBytes)
```
chhtz, line 35: I would make `val` an `int`/`int32_t` in case someone wants to initialize a Frame with `depth=16` by `0xffff` (may not happen often, but won't cost anything)
(same for other occurrences of `val`)

src/samples/Frame.hpp
```diff
@@ -12,272 +12,298 @@
...
+         *     it is computed from the other parameters
+         */
+        void init(uint16_t width, uint16_t height, uint8_t depth=8, frame_mode_t mode=MODE_GRAYSCALE, int16_t val = 0, size_t sizeInBytes=0);
+
+        /** Fill the image buffer's bytes by a given value
+         *
+         * @param val a value used to initialize each byte of the image's buffer.
+         *     if negative, it is left uninitialized
+         */
+        void reset(int val = 0);
```
chhtz, line 118: constness of `val` should be removed in the cpp-file as well.

The current implementation of `reset` assumes 8-bit images (for 16bit images the upper and lower byte will be set equal). It would be possible to generalize this using `std::fill_n` (not sure if this is needed, probably the only actual use-cases are `0` and `0xff`)
Instead of changing the implementation, I'd also be fine with documenting the current behavior)

chhtz (changes requested):

src/samples/Frame.cpp
doudou: The code currently does not handle that case (it's simply filling the vector which is a int8_t)

src/samples/Frame.hpp
doudou: That's of course right ... but then it built on my machine ... hum

doudou: > Instead of changing the implementation, I'd also be fine with documenting the current behavior)

That's why I wrote "each byte of the image buffer". But I'll be more explicit.

doudou: Updated

chhtz: Actually, C++ appears to ignore const-ness of POD parameters (except inside the function body of course). And different const-ness in declaration and implementation does not throw warnings (not even with `-Weverything` on clang), so allowing this seems to be intentional (https://godbolt.org/z/bGPPT8hT3).

chhtz (approved):

wvmcastro (approved):

src/samples/Frame.hpp
pierrewillenbrockdfki: Making a by-value parameter const does not change anything outside of the function itself, and in c++, all parameters are by-value(references and pointers are not special, the address is passed by-value). Another thing that is only relevant to the signature preceding the implementation are parameter names, and as such, you can have different names in forward declarations, or have none at all. Especially different names are confusing to no end.

For consistency and sanity, it is probably good to keep the signatures in synch.

pierrewillenbrockdfki (approved):
Looks good to me, Thanks for keeping the (mostly) whitespace changes in a separate commit from the functional ones.

PR #173, opened 2025-04-09
Add serialization for Angle.hpp

haider8645:
Hello @planthaber,

I had to make the [TravNode.hpp](https://github.com/dfki-ric/traversability_generator3d/blob/5f5ae5278a1ad16ea67a1c68159813c21f4b0292/src/TravGenNode.hpp#L31) available over a rock port and it needs the Angle.hpp to be serialized.

Best,
Haider

### DIFF
--- a/src/Angle.hpp
+++ b/src/Angle.hpp
@@ -7,6 +7,7 @@
 #include <base/Eigen.hpp>
 
 #include <vector>
+#include <boost/serialization/serialization.hpp>
 
 namespace base
 {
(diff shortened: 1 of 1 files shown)


chhtz:
Not sure if we need to pull in boost-serialization into this. Probably an [out-of-class definition](https://www.boost.org/doc/libs/1_87_0/libs/serialization/doc/serialization.html#free) inside TravNode.hpp would be a less intrusive option?

But I don't strictly oppose adding it here.

haider8645:
> Not sure if we need to pull boost-serialization into this. Probably an [out-of-class definition](https://www.boost.org/doc/libs/1_87_0/libs/serialization/doc/serialization.html#free) inside TravNode.hpp would be a less intrusive option?
> 
> But I don't strictly oppose adding it here.

Thanks, I'll do that.

haider8645:
@chhtz the out of class definiton works. Thanks!

I'll close this MR, alright?

haider8645:
> Unrelated: Why does AngleSegment have a width a start and an end? The end appears to be calculated always in the constructor and I doubt that caching an addition result is worth adding a member variable ...

There is a function getEnd() and endRad is used in various computations. Isn't it worth to have it saved as a member rather than always computing from start and width?

PR #174, opened 2025-04-09
chore: include meta flag for logical time in base samples

kapeps:
After this PR https://github.com/rock-core/tools-syskit/pull/452 
It is possible to signal to use the base::Time as logical time.

This PR, signals all the time fields in the base/samples that are supposed to be used as logical time when logging.

Obs: I tried to follow the same documentation style for each file... and tested that all of these styles flagas are detected.

### DIFF
--- a/src/samples/Acceleration.hpp
+++ b/src/samples/Acceleration.hpp
@@ -9,6 +9,7 @@ namespace base { namespace samples {
 /** Spatial Acceleration with sampled time and frame ID*/
 struct Acceleration : public base::Acceleration
 {
+    /** @meta role logical_time */
     base::Time time;
     /** ID of the coordinate frame in which the spatial acceleration is expressed*/
     std::string frame_id;
--- a/src/samples/BodyState.hpp
+++ b/src/samples/BodyState.hpp
@@ -36,7 +36,10 @@ namespace base { namespace samples {
         BodyState(const base::TransformWithCovariance& pose, const base::TwistWithCovariance& velocity):
             pose(pose), velocity(velocity) {};
 
-        /** Time-stamp **/
+        /**
+         * Time-stamp
+         *
+         * @meta role logical_time */
         base::Time time;
 
         /** Robot pose: rotation in radians and translation in meters */
--- a/src/samples/BoundingBox.hpp
+++ b/src/samples/BoundingBox.hpp
@@ -62,7 +62,10 @@ struct BoundingBox {
      */
     bool hasValidCovDimension() const;
 
-    /* The sample timestamp. */
+    /**
+     * The sample timestamp.
+     *
+     * @meta role logical_time */
     Time time;
 
     /* 3D center (x, y, z) position of the bounding box . */
--- a/src/samples/DepthMap.hpp
+++ b/src/samples/DepthMap.hpp
@@ -61,7 +61,8 @@ public:
      * This timestamp is used for temporal alignment to other data samples
      * and transformations. 
      * It is important to always set here a meaningful value.
-     */
+     *
+     * @meta role logical_time */
     base::Time time;
     
     /** The timestamps can be either one timestamp for all measurements,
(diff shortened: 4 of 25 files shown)


chhtz:
Would it be possible to make `.time` the default "logical time" for every sample-type and only require a flag if a different field shall be used? (I don't fully understand the linked PR nor for what functionality it is actually required)

I think it is a bit dangerous to depend on a flag for which it is not obvious if/why it is needed (nothing would break immediately if one accidentally removes any of these flags, I assume?)

doudou:
I don't think that's a very good idea. There are types (i.e. command types) for which `time` is not so well defined and is in effect not the type's logical time.

I prefer having to explicitly think about the problem and mark the field.

doudou:
>  (I don't fully understand the linked PR nor for what functionality it is actually required)

It is the field that is passed to the logger to fill the log file's logical time.

chhtz:
> It is the field that is passed to the logger to fill the log file's logical time.

Ok, then what shall be the logical time for types which don't have a `time` member, and would it be easier to just mark these types? Or is there a way to mark a field for types which I don't define myself

I think my main concern is to add something which is hard to verify that it is done correctly. Also, if for some reason these tags are not used anymore in the future, they likely will still exist and it is hard to know whether these are mandatory, optional or obsolete (and probably nobody dares to remove them) ...

Overall, I'm not opposing to add this, as just adding comments won't break any existing code ...

doudou:
> Ok, then what shall be the logical time for types which don't have a time member,

Such a type does not have a logical time field, logging will use realtime for that. This is the current situation. Adding the comments improves on it, it's not breaking to not have the comment (you continue with the current situation)

> and would it be easier to just mark these types?

I don't think so. There is already a lot of types falling in the "there is a base::Time but it's not the logical time". Namely, most command types.

Generally speaking, I prefer to have a slightly wrong logical time ("realtime") than a very wrong one. I think the convention of picking the base::Time field would lead to the latter.

> Or is there a way to mark a field for types which I don't define myself?

Not right now ... but it wouldn't be so hard to implement actually.

> as just adding comments won't break any existing code ...

Exactly. It is also documentation, as it makes it clear that this field is the type's logical time field.

wvmcastro (approved):
The only thing that bothers me is not having any documentation saying 
"this is not an ordinary comment". Although I don't know where would be the best place to put it because it is actually a syskit feature but enabled in other packages

doudou:
> Although I don't know where would be the best place to put it because it is actually a syskit feature but enabled in other packages

It's a convention that is currently only used by the syskit-based toolchain.

`@meta` is parsed by typelib and stored in the types' metadata. Any toolchain that uses typelib (which is most probably anything that uses orogen in the end) has access to the info.

Again, note that types that do not have the annotation won't be affected.

As for @chhtz's concern about changing the meaning of a field ... I would be very concerned if people do that (i.e. if people start using a time field whose interpretation was the data's logical time to suddently not be it ...)

doudou:
> Ok, then what shall be the logical time for types which don't have a `time` member, and would it be easier to just mark these types? Or is there a way to mark a field for types which I don't define myself?
> 
> I think my main concern is to add something which is hard to verify that it is done correctly. Also, if for some reason these tags are not used anymore in the future, they likely will still exist and it is hard to know whether these are mandatory, optional or obsolete (and probably nobody dares to remove them) ...
> 
> Overall, I'm not opposing to add this, as just adding comments won't break any existing code ...

Re-reading this, I actually don't fully understand your concerns.

This is machine-readable documentation about the interpretation of the time field. The issue with just picking "time" is that some "time" fields are not the sample's logical time.

Relying on that information is relying on documentation. Of course, some people might not fill the data as it should, but this is orthogonal to having the annotation (RBS expression frame of the angular velocity *cough*)

As for removing the flag, I'm not sure about the concern either. As I said, it is documentation.

PR #176, opened 2025-09-02
fix: use bin count getter at sonar visualization

seixasxbr:
fix: remove division by bin_count in fan segment calculus
the bin_count inside a Sonar struct is the number of bins in each beam
image

### DIFF
--- a/src/samples/Sonar.hpp
+++ b/src/samples/Sonar.hpp
@@ -116,8 +116,10 @@ public:
     
     /**
      * Number of bins in a beam
-     * 
-     * @deprecated, use getBeamBinCount() to return the number of bins in each beam
+     *
+     * @deprecated this field seem to be filled inconsistently across packages,
+     * and actually does not provide useful information. Use \c
+     * getSingleBeamBinCount() instead
      */
     uint32_t bin_count;
     
@@ -288,7 +290,7 @@ public:
      * Returns the number of bins in each beam
      * 
      */
-    uint32_t getBeamBinCount();
+    uint32_t getSingleBeamBinCount() const;
 
 BASE_TYPES_DEPRECATED_SUPPRESS_START
     explicit Sonar(SonarScan const& old, float gain = 1);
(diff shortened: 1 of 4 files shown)


chhtz:
I think generally that makes sense, but this may require fixing any existing sonar driver (if they used `bin_count` to make the visualization look correct, instead of according to its documentation ...)

src/samples/Sonar.hpp
```diff
@@ -114,7 +114,11 @@ struct Sonar
     /** The speed of sound in water at the time of acquisition and in m/s*/
     float speed_of_sound;
     
-    /** Number of bins in a beam */
+    /**
+     * Number of bins in a beam
+     * 
+     * @deprecated, use getBeamBinCount() to return the number of bins in each beam
```
doudou, line 120: `@deprecated this field seem to be filled inconsistently across packages, and actually does not provide useful information. Use \c getSingleBeamBinCount() instead`

src/samples/Sonar.hpp
```diff
@@ -280,6 +284,12 @@ struct Sonar
      */
     void validate();
 
+    /**
+     * Returns the number of bins in each beam
+     * 
+     */
+    uint32_t getBeamBinCount();
```
doudou, line 291: Since we are trying to be very unambiguous ...

```
/** The number of bins in a single beam
 */
uint32_t getSingleBeamBinCount() const;
```

(Note the extra "const")

test/viz/test_sonar_visualization.rb
```diff
@@ -20,7 +20,7 @@ def init_sonar_sample(beam_count, bin_count, step, initial_bearing = 0)
             sonar = Types.base.samples.Sonar.new
             sonar.speed_of_sound = 1000
             sonar.beam_count = beam_count
-            sonar.bin_count = beam_count*bin_count
```
doudou, line 23: Remove this change.

doudou:
This is not creating any backward-compatibility issues anymore ... merging

PR #178, opened 2026-01-07
Compatibility with rice version 4.x

chhtz:
Not sure if that is the correct way to handle compatibility with rice 4.x


bindings/ruby/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ else(SISL_FOUND)
         EXT_RICE base_types_ruby Eigen.cpp init.cpp)
 endif(SISL_FOUND)
 
+target_compile_features(base_types_ruby PUBLIC cxx_std_17)
```
doudou, line 12: This is base-types ... I would avoid doing that blindly.

If the goal is to support rice > 4, could you do it conditionally.

chhtz: By "conditionally", do you mean checking the rice version, or adding a cmake flag?
I can try either. I think I had problems finding out the rice version (by cmake), but it has been a while (I had this change as local fix some time before I made this PR)

PR #179, opened 2026-02-13
chore: remove our modified version of boost/circular_buffer.hpp

doudou:
This removes compatibility with Boost < 1.62.0. Given that this was released in
2016, I think that's fair enough.

Change that fixed the reason of CircularBuffer.hpp's existence:

https://github.com/boostorg/circular_buffer/commit/ea60799f315aa2e861d0e14ca9012950021c2fc6


pierrewillenbrockdfki (approved):
This only affects ubuntu before 18.04 (which is at 1.65).

chhtz (approved):

jhonasiv (approved):

